PR #5, opened 2016-02-28
Scala API

rouzwawi:
This is a Scala version of the [TaskBuilder](http://rouz.io/flo/maven/apidocs/io/rouz/task/dsl/TaskBuilder.html) API. It is implemented as a wrapper to the existing java based API in order to keep the implementation complexity to a minimum.

### DIFF
--- a/flo-scala/pom.xml
+++ b/flo-scala/pom.xml
@@ -8,9 +8,10 @@
   </parent>
 
   <name>Flo Scala API</name>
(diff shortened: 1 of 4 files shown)


nevillelyh:
LGTM

At some point we should machine generate overloaded methods with type params `[A, B, C, ...]`?

rouzwawi:
Yes, I think that should be done for both the java and the scala api.

PR #16, opened 2016-05-27
Add currentTaskId method to TaskContext

### LINKED ISSUE #15: A way to get current TaskId from TaskContext
It's useful to know which `TaskId` is being executed currently from inside `processWithContext` calls.

rouzwawi:
This enabled task utilities to give the user useful information about the task being evaluated. The method is intended to be used from the functions passed to any of the `processWithContext(...)` builders.

fixes #15

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,12 @@
             <version>1.3</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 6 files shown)


rouzwawi:
@nevillelyh

PR #28, opened 2017-09-06
Task Metadata

rouzwawi:
This PR adds support for adding arbitrary, type-safe metadata to tasks. The API is inspired by Kotlins [Coroutine Context Elements](https://github.com/Kotlin/kotlin-coroutines/blob/master/kotlin-coroutines-informal.md#coroutine-context).

Metadata elements are added to tasks keyed on a key type that is parameterized on the value type.

```java
interface MetaKey<E>
```

Metadata key-values can be added to a task through the following builder method.

```java
<E> TaskBuilderX meta(MetaKey<E> key, E value)
```

Added values can be accessed from the `Task` object through a type-safe accessor method.

```java
<E> Optional<E> getMeta(MetaKey<E> key)
```

## Metadata from Operators

Operators can add metadata to a task by supplying a list of entries to the task they are being added to.

```java
List<MetaKey.Entry<?>> provideMeta()
```

In order to make the construction of this list type-safe, `MetaKey<E>` has a method for constructing an `Entry<E>`.

```java
class MyOperator extends OpProvider<String> {

  private static final MetaKey<String> TEST_KEY_1 = MetaKey.create("key1");
  private static final MetaKey<Integer> TEST_KEY_2 = MetaKey.create("key2");

  // ...

  @Override
  public List<MetaKey.Entry<?>> provideMeta() {
    return Arrays.asList(
        TEST_KEY_1.value("foobar"),
        TEST_KEY_2.value(42)
    );
  }
}
```

### DIFF
--- a/flo-api-generator/src/main/resources/ScalaApi.mustache
+++ b/flo-api-generator/src/main/resources/ScalaApi.mustache
@@ -7,7 +7,7 @@ import com.spotify.flo._
 import scala.reflect.ClassTag
 
 object FloTask {
-  def named[Z: ClassTag](name: String, args: Any*): TaskBuilder0[Z] = new Builder0(name, args:_*)
+  def named[Z: ClassTag](name: String, args: Any*): TaskBuilder0[Z] = Builder0(name, args:_*)
 }
 
 trait TaskBuilder0[Z] {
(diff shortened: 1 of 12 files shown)


danielnorberg:
What was the intended use case for this feature?

honnix:
Let's close this for now @danielnorberg ?

PR #29, opened 2017-09-12
OpProvider interface requires type of Task's evaluated obj

fabriziodemaria:
This will make `onSuccess` safer when the operator expects a well defined type for the returned object from the Task. However, generic operators that would act on `onSuccess` without caring about the returned object are not supported as good anymore. We can have two types of operators, the generic one with a different signature for the `onSuccess` method (i.e. no second argument `z`).


rouzwawi:
I will 👎  this because it conflates the operator with the task return type. That would be an impediment to using several operators with a task as they would now have to use a common type.

For the use cases this change was made for, the best way to approach it would be to pass in the required information to the operator while constructing it instead of relying on the task return type.

PR #33, opened 2017-09-26
[WIP] Add OpProviderStrict

fabriziodemaria:
This makes #29 and #31 obsolete.

Still figuring out how to pass the result value for non-strict OpProvider's `onSuccess` method. Perhaps the best solution is to not give access at all to the result value if the OpProvider is not strict, thus avoiding to operate on generic Object altogether.


fabriziodemaria:
Closing in favour of #37

PR #37, opened 2017-09-29
Add OpProviderStrict and OpProviderGeneric

fabriziodemaria:
Makes `OpProvider` a basic interface not to be exposed to providers' implementations. The choice is now between `OpProviderGeneric` (not tied to the task's return type) and `OpProviderStrict` (tied to the task's return type).


fabriziodemaria:
I am actually not sure I agree we need to make `onSuccessStrict` abstract. There are cases where the Provider's author needs type safety with respect to the returned value of the Task without the need to execute anything in case of Task's success. After all, for `OpProviderGeneric` the `onSuccess` method is not abstract either.

fabriziodemaria:
@jocke third opinion on the `abstract` thingy?

bergman:
I don't see the reason why it has to be abstract in the strict case if the generic one is non-abstract. It should be both or either.

fabriziodemaria:
> I don't see the reason why it has to be abstract in the strict case if the generic one is non-abstract. It should be both or either.

Agree. Removing `abstract` and merging 👍

PR #40, opened 2017-10-09
use utf-8 instead of system default

honnix:
@fabriziodemaria PTAL

This is NOT necessarily needed to get correct print out.


fabriziodemaria:
👍

PR #44, opened 2017-10-10
Fail the task if memoizer.store fails

honnix:
By creating a new Value.

### DIFF
--- a/flo-workflow/src/main/java/com/spotify/flo/context/MemoizingContext.java
+++ b/flo-workflow/src/main/java/com/spotify/flo/context/MemoizingContext.java
@@ -22,7 +22,6 @@ package com.spotify.flo.context;
 
 import static com.spotify.flo.Util.colored;
 
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.spotify.flo.EvalContext;
@@ -268,7 +267,7 @@ public class MemoizingContext extends ForwardingEvalContext {
     try {
       return method.invoke(/* static */ null, args);
     } catch (IllegalAccessException | InvocationTargetException e) {
-      throw Throwables.propagate(e);
+      throw new RuntimeException(e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


honnix:
There is an issue though. If we have a very deep graph (deeper than number of workers), we will have deadlock, basically all workers are used and blocked.

honnix:
I might have a better solution.

rouzwawi:
Can you provide a test that reproduces the problem so we know it's fixed.

honnix:
Need to think again. This can create a deadlock and haven't found a good solution yet. @rouzwawi

rouzwawi:
Does the deadlock happen with AsyncContext + Memoizing or with the SyncContext? If the former, it's surprising, that why I want to se a test that reproduces it so we can be sure.

honnix:
It’s the former. Basically this PR is broken because it blocks the executor thread. I found the issue when working on a test case actually: async with single thread pool. I can commit the test case later.

rouzwawi:
Oh ok. I thought that the PR was supposed to fix the the deadlock issue, but it's actually causing it?

honnix:
Yeah unfortunately.

honnix:
The purpose of the PR is to fail a task if memoizer.store fails. And the idea is somehow to hijack the actual function. But doesn’t work nicely due to the blocking implementation.

rouzwawi:
Could it work through `flatMap`ping the Value and failing it in there?

honnix:
Could be. I started that track before heading home. Will try later. Thanks for taking a look at this!

honnix:
I’m more sure that’s gonna work now, reading flatten.

honnix:
@rouzwawi Maybe just map.

rouzwawi:
Could work too

flo-workflow/src/main/java/com/spotify/flo/context/MemoizingContext.java
```diff
@@ -215,10 +215,12 @@ private void mapMemoizer(Memoizer<?> memoizer) {
     final Task<T> task = evalBundle.task;
     final Memoizer<T> memoizer = evalBundle.memoizer;
 
-    final Value<T> tValue = delegate.invokeProcessFn(taskId, processFn);
-    tValue.consume(v -> memoizer.store(task, v));
+    final Value<T> newValue = processFn.get().map(v -> {
```
rouzwawi: Won’t this call to `processFn.get()` change when it’s called? From being decided by the delegate to being called here. I would best this code inside a `() -> ...` block so it can be passed into the delegate directly.

honnix: Yeah, good point.

honnix:
Somehow my repo has become a mess...

PR #45, opened 2017-10-10
configure it default to be utf-8 for non-verbose logging

honnix:
@fabriziodemaria PTAL Then we don't need to set `file.encoding`.


fabriziodemaria:
Where do we set `file.encoding` now?

honnix:
No need to set if we merge this.

fabriziodemaria:
👍

bergman:
what does `non-verbose` mean in the title of this PR?

honnix:
https://github.com/spotify/flo/blob/master/flo-runner/src/main/java/com/spotify/flo/context/ConsoleLogging.java#L154

If it is verbose, we configure the layout using code, which we have already fixed by setting proper charset.

PR #47, opened 2017-10-10
Fail the task if `memoizer.store` fails.

honnix:
A new PR to fix commit mess of #44

### DIFF
--- a/flo-workflow/src/main/java/com/spotify/flo/context/MemoizingContext.java
+++ b/flo-workflow/src/main/java/com/spotify/flo/context/MemoizingContext.java
@@ -108,8 +108,8 @@ public class MemoizingContext extends ForwardingEvalContext {
     @interface Impl { }
 
     /**
-     * Lookup a memoized value for a given task. Implementation can throw out a {@link RuntimeException}
-     * and that would fail the task evaluation.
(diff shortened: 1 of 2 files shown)


honnix:
@rouzwawi I closed #44 as my repo was in a very weird state.

fabriziodemaria:
@honnix are your concerns about deadlock still valid in this new PR?

honnix:
No. This is a much better solution. Basically constructing a new `Value`.

fabriziodemaria:
I see, nice! Let's update the JavaDocs with the new behaviour of `store`.

We should also improve the printed logs in case of errors to make clear which Task failed publishing.

honnix:
There is no new behaviour of `store`. We might want to make it explicitly declare throwing a checked exception, or in Javadoc make it clear what if the implementation throws `RuntimeException`.

honnix:
In terms of logging, we are actually printing task id. E.g.
```
11:04:47.229 flo-worker-0                 | WARN | FloRunner  |> streamCount1(2016-10-01T00)#f5f57287 Failed
java.lang.RuntimeException: forced failure
...
```

fabriziodemaria:
👍

PR #48, opened 2017-10-11
Make Listener interface closeable

NatashaL:


flo-runner/src/main/java/com/spotify/flo/context/ChainedListener.java
```diff
@@ -61,4 +64,29 @@ private void guardedCall(Runnable call) {
       logging.exception(t);
     }
   }
+
+  @Override
+  public void close() throws IOException {
```
fabriziodemaria: Can we add a test around this?

flo-runner/src/main/java/com/spotify/flo/context/ChainedListener.java
```diff
@@ -61,4 +64,29 @@ private void guardedCall(Runnable call) {
...
+    final Optional<IOException> exception =
+        Stream.of(first, second)
+            .map(this::close)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .findFirst();
+
+    // TODO: Throw a composition of two exceptions when both first and second throw exceptions
```
fabriziodemaria: Is it intended to do this in another PR?

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
```diff
@@ -49,7 +50,7 @@
   /**
    * A listener for instrumented evaluation. See {@link InstrumentedContext} for more details.
    */
-  public interface Listener {
+  public interface Listener extends Closeable {
```
honnix: Having a default implementation in `Listener` which does nothing?

NatashaL: We can only override the declaration of the `close()` method, not provide an empty implementation, due to Listener being an interface itself. But it seems unnecessary. Whoever implements `Listener` in their implementation class they will be required to implement the `close()` method too.

flo-runner/src/main/java/com/spotify/flo/context/ChainedListener.java
NatashaL: Yes, see 3e7083f

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
honnix: That you can have for an interface.

NatashaL: Right, yes, we can do that.

bergman:
squashed and merged in 0600752

PR #51, opened 2018-01-24
stop depending on logback

bergman:
- Remove ability to configure logging verbosity from flo config and instead shifts that responsibility onto the logback configuration in the user's project

One drawback of this is that if you don't include logback in your project that uses flo, you will get no output at all. If you don't have a logback.xml in your project but do depend on logback, you will get the default configuration that will print a lot of things. These drawbacks are common in any library. For example, the logback.xml that was here before had [configuration](https://github.com/spotify/flo/blob/555a8a84ad89767f8692628dba0ae1bf122282a9/flo-runner/src/main/resources/logback.xml#L14) to make apollo not log anything below WARN level.


bergman:
One drawback of this is that if you don't include logback in your project that uses flo, you will get no output at all.

ulzha:
:+1: makes a lot of sense.

PR #53, opened 2018-02-06
FloRunner refactor

bergman:
- make FloRunner easier to test
- return a useful `Result` object rather than call `System.exit` (replaced by `FloRunner.runTask(foo).waitAndExit()`)
- remove the args to task and the `TaskConstructor` interface which was used mostly for CLI usage
- no more apollo
- no more guava
- remove json tree logging


bergman:
It would be cool to be able to get the value of the root task as a `Future` in the `Result` object so I might look into that.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -20,116 +20,113 @@
...
+      try {
+        close();
+      } catch (Exception e) {
+        e.printStackTrace();
+      }
+    });
+
+    value.onFail(t -> {
+      logging.exception(t);
+      final long elapsed = System.currentTimeMillis() - t0;
```
bergman: This code is identical to the `consume` handler above… I couldn't find a way to handle it in one place :/

flo-workflow/src/main/java/com/spotify/flo/context/MemoizingContext.java
```diff
@@ -164,7 +165,7 @@ public static Builder builder(EvalContext baseContext) {
   public static class Builder {
 
     private final EvalContext baseContext;
-    private final ImmutableMap.Builder<Class<?>, Memoizer<?>> memoizers = ImmutableMap.builder();
+    private final Map<Class<?>, Memoizer<?>> memoizers = new HashMap<>();
```
bergman: This should be fine, right?

fabriziodemaria:
Overall looks good to me.
Regarding the new asynchronous `runTask()`, we could rename it `runTaskAsync()`. `runTask()` could be a blocking version that still returns the resulting value and re-throws any exception from Flo or from within the Tasks.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
bergman: found a way :)

bergman:
I opted for the `Result` type anyway. The upgrade case will be a bit painful but assuming not a lot of people use this yet, it should be fine.

bergman:
> Overall looks good to me.
Regarding the new asynchronous `runTask()`, we could rename it `runTaskAsync()`. `runTask()` could be a blocking version that still returns the resulting value and re-throws any exception from Flo or from within the Tasks.

I opted for breaking the API. The motivation is that it is still quite early in the project and to me the async should be default.

rouzwawi:
LGTM :+1:

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -244,4 +217,51 @@ public static String randomAlphaNumeric(int count) {
...
   }
+
+  public static class Result<T> {
+    private final Future<T> future;
+
+    Result(Future<T> future) {
+      this.future = future;
+    }
+
+    public Future<T> future() {
```
danielnorberg: Why `Future` (which only allows blocking `get()`) instead of `CompletableFuture` or `CompletionStage` ?

PR #54, opened 2018-02-13
add input/inputs to task builder

bergman:


fabriziodemaria:
Should we just remove `in` to avoid further confusion in the future, despite breaking change?
In any case let's add two simple tests with the new name.

bergman:
I would remove all the graphic/ascii-art aliases as well then. I will add a test that uses these new names at least.

bergman:
all the ascii art method names have been dropped now and the tests all refer to `input` and `inputs`.

PR #57, opened 2018-03-11
allow memoizer to mutate value on store

bergman:


bergman:
Found another way to do this so not needed at this point.

Issue #64, opened 2018-05-02
Upgrade to mockito 2


thomasdziedzic-pd:
I got an error with conflicting classes when I included flo-scala into my project.

It seems that mockito-all is deprecated since 2015 and should probably be removed or replaced.

See https://github.com/mockito/mockito/issues/324 & https://github.com/mockito/mockito/commit/d34597c33d3a5a3971ff1db68ef1cb8a1145ac03

PR #65, opened 2018-05-02
remove mockito-all dependency

thomasdziedzic-pd:
This fixes part of the problem with https://github.com/spotify/flo/issues/64

It seems that mockito-all isn't used anywhere? CircleCI tests aren't running automatically, but I ran `mvn verify` locally with my changes and it seems like all tests are passing.


labianchin:
+1

PTAL @bergman

bergman:
Thanks!

PR #66, opened 2018-05-03
removed duplicate dependency on mockito and fixed its scope

balaji:
This also fixes an issue where `mockito` libraries causes dependency conflicts in clients.


bergman:
Closing since #65 was merged

PR #67, opened 2018-05-03
add AsyncContext example

thomasdziedzic-pd:
I had a hard time trying to get an AsyncContext correctly working.
Added an example so that others can use it for their own use case.


rouzwawi:
Thanks for the example. Looks great!

PR #68, opened 2018-05-08
[Option 1] Add TerminationHook and FloRunner.Builder

NatashaL:


flo-runner/src/main/resources/reference.conf
```diff
@@ -11,3 +11,5 @@ styx.workflow.parameter = "UNKNOWN_PARAMETER"
 styx.workflow.parameter = ${?STYX_PARAMETER}
 styx.workflow.execution = "UNKNOWN_EXECUTION"
 styx.workflow.execution = ${?STYX_EXECUTION_ID}
```
honnix: Heh, didn't recall we have this already. So we might just make our life easier by removing the internal code and reading environment variables.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -57,16 +57,56 @@
...
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static FloRunner createDefault() {
+    return newBuilder().build();
+  }
 
   private FloRunner(Config config) {
```
honnix: We have the opportunity to break backward compatibility, so let's break it badly.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -57,16 +57,56 @@
 /**
  * This class implements a top-level runner for {@link Task}s.
  */
-public final class FloRunner<T> {
+public class FloRunner<T> {
 
   private static final Logger LOG = LoggerFactory.getLogger(FloRunner.class);
+  private static final TerminationHookFactory DEFAULT_TERMINATION_HOOK_FACTORY = () ->
```
honnix: It doesn't feel we need a factory. A default hook does nothing but exists should be sufficient.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,6 +121,10 @@ private static Config defaultConfig() {
    * @return a {@link Result} with the value and throwable (if thrown)
    */
   public static <T> Result<T> runTask(Task<T> task, Config config) {
+    if (config.getBoolean("termination.logging")) {
```
honnix: We could make this method non-static. Would that be easier for the internal part?

NatashaL:
Opted for https://github.com/spotify/flo/pull/69 instead

PR #69, opened 2018-05-08
Add termination-hook capability

NatashaL:
This alternative uses a service loader to load the injected dependencies.


flo-runner/src/main/resources/reference.conf
```diff
@@ -11,3 +11,6 @@ styx.workflow.parameter = "UNKNOWN_PARAMETER"
 styx.workflow.parameter = ${?STYX_PARAMETER}
 styx.workflow.execution = "UNKNOWN_EXECUTION"
 styx.workflow.execution = ${?STYX_EXECUTION_ID}
+
+# log exit code to a file
+termination.logging = true
```
honnix: We probably don't need this flag, `loadTerminationHook` can return a default hook if nothing loaded.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,9 +83,29 @@ private static Config defaultConfig() {
...
     return new Result<>(new FloRunner<T>(config).run(task));
   }
 
+  private static Optional<TerminationHook> loadTerminationHook() {
+    final ServiceLoader<TerminationHookFactory> factories =
+        ServiceLoader.load(TerminationHookFactory.class);
+
+    final Iterator<TerminationHookFactory> factoryIterator = factories.iterator();
+    if (factoryIterator.hasNext()) {
+      return Optional.of(requireNonNull(factoryIterator.next().create()));
```
honnix: `Optional.of` does `requireNonNull` by itself.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,9 +83,24 @@ private static Config defaultConfig() {
    * @return a {@link Result} with the value and throwable (if thrown)
    */
   public static <T> Result<T> runTask(Task<T> task, Config config) {
+    Optional<TerminationHook> terminationHook = loadTerminationHook();
+    if (terminationHook.isPresent()) {
+      return new Result<>(new FloRunner<T>(config).run(task), terminationHook.get());
```
honnix: Not important. How do you think having one constructor for Result and here we default to System:exit? So it would become terminationHook.getOrElse here and we don’t need branches.

NatashaL: I agree

NatashaL:
I'll also try to make the termination hooks chained. Assuming we want to enforce always exiting with the exit code of the task as is here https://github.com/spotify/flo/pull/69/files#diff-a67d9bdd9c0d9f83c3aa2abead4b570bR289 .

So far, with this implementation, we trust the overriding TerminationHook to also call System.exit(exitCode), which might be forgotten and cause inconsistencies between what's logged and what the process shows.

@honnix  thoughts?

honnix:
Sounds reasonable. I would even make assume the hook will never exit but just run something. After calling all hooks, we exit.

honnix:
Or manually append exit hook to the end.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -234,21 +248,28 @@ public static String randomAlphaNumeric(int count) {
 
   public static class Result<T> {
     private final Future<T> future;
+    private final Consumer<Integer> exiter;
 
-    Result(Future<T> future) {
+    Result(Future<T> future, Consumer<Integer> exiter) {
       this.future = future;
+      this.exiter = (Integer exitCode) -> {
```
honnix: Is this type declaration needed?

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -234,21 +248,28 @@ public static String randomAlphaNumeric(int count) {
...
+        exiter.accept(exitCode);
```
honnix: We need to catch and ignore exception here. If javaslang is already a dependency we can use Try.of.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -234,21 +248,28 @@ public static String randomAlphaNumeric(int count) {
 
   public static class Result<T> {
     private final Future<T> future;
+    private final Consumer<Integer> exiter;
 
-    Result(Future<T> future) {
+    Result(Future<T> future, Consumer<Integer> exiter) {
```
honnix: If we do it this way, we might just pass in an optional of hook and run it if it is not empty, followed by System.exit, so we don’t need to write that code twice and no need for the comment either. Or even better just pass in an iterator of optionals (of hooks) and loop through it, so we can keep the chain.

honnix: Also I guess we want to use TerminationHook as argument type?

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -234,21 +247,26 @@ public static String randomAlphaNumeric(int count) {
 
   public static class Result<T> {
     private final Future<T> future;
+    private final Consumer<Integer> exiter;
 
-    Result(Future<T> future) {
+    Result(Future<T> future, Iterable<TerminationHook> terminationHooks) {
       this.future = future;
+      this.exiter = (Integer exitCode) -> {
+        terminationHooks.forEach(hook -> hook.accept(exitCode));
```
honnix: try/catch hook execution.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,7 +85,16 @@ private static Config defaultConfig() {
...
   public static <T> Result<T> runTask(Task<T> task, Config config) {
-    return new Result<>(new FloRunner<T>(config).run(task));
+    return new Result<>(new FloRunner<T>(config).run(task), loadTerminationHooks());
+  }
+
+  private static Iterable<TerminationHook> loadTerminationHooks() {
+    final ServiceLoader<TerminationHookFactory> factories =
+        ServiceLoader.load(TerminationHookFactory.class);
+
+    return StreamSupport
```
honnix: Maybe just do it in old school way: passing in `iterator` and later do `iterator.forEachRemaining`.

honnix: nvm, it is a iterator of factories. makes sense to fail the hook instance creation early on.

pom.xml
```diff
@@ -150,6 +150,18 @@
...
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito</artifactId>
+      <version>1.7.4</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
```
honnix: Using power mock usually speaks for a design smell. We have been trying to avoid it I believe.

NatashaL: hmm, what would be a good way to test the dynamically loaded classes in that case?

honnix: Implement the interface (probably just a lambda) in test package. In the hook set a flag?

NatashaL: The problem is actually loading the implementation class from the test package, so that the ServiceLoader recognizes it. I can't find a straightforward way to do it. It's indeed an indicator that this may not be the best approach.

NatashaL: nvm, found what I was missing, albeit might not be the best still

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,7 +86,27 @@ private static Config defaultConfig() {
...
+        .stream(
+            Spliterators.spliteratorUnknownSize(factories.iterator(), Spliterator.ORDERED), false)
+        .filter(Objects::nonNull)
+        .map(factory -> {
+          try {
```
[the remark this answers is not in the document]
honnix: `create` doesn't throw checked exception, why bothering catching and wrapping? let's just remove try/catch here.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,7 +86,26 @@ private static Config defaultConfig() {
...
+            return factory.create();
+          } catch (Exception e) {
+            throw new RuntimeException(e);
+          }
+        })
+        .filter(Objects::nonNull)
```
honnix: This also hides potential issue. Let it fail instead of swallowing? I mean we don't need to check null.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,7 +86,26 @@ private static Config defaultConfig() {
...
+  }
+
+  private static Iterable<TerminationHook> loadTerminationHooks() {
+    final ServiceLoader<TerminationHookFactory> factories =
+        ServiceLoader.load(TerminationHookFactory.class);
+
+    return StreamSupport
+        .stream(
+            Spliterators.spliteratorUnknownSize(factories.iterator(), Spliterator.ORDERED), false)
+        .filter(Objects::nonNull)
```
honnix: Why would we need this?

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -234,21 +258,25 @@ public static String randomAlphaNumeric(int count) {
...
   public static class Result<T> {
     private final Future<T> future;
+    private final Iterable<TerminationHook> terminationHooks;
+    private Consumer<Integer> exiter;
 
-    Result(Future<T> future) {
+    Result(Future<T> future, Iterable<TerminationHook> terminationHooks) {
       this.future = future;
+      this.terminationHooks = terminationHooks;
+      this.exiter = System::exit;
```
honnix: This field doesn't seem to be necessary now after a few refactoring.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -260,9 +288,10 @@ public T value() throws ExecutionException, InterruptedException {
 
     // visible for testing
     void waitAndExit(Consumer<Integer> exiter) {
+      this.exiter = exiter;
```
honnix: Pass in `exiter` to `exit` instead of setting to a field?

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -81,7 +86,26 @@ private static Config defaultConfig() {
...
+    final ServiceLoader<TerminationHookFactory> factories =
+        ServiceLoader.load(TerminationHookFactory.class);
+
+    return StreamSupport
+        .stream(
+            Spliterators.spliteratorUnknownSize(factories.iterator(), Spliterator.ORDERED), false)
+        .filter(Objects::nonNull)
+        .map(factory -> {
+          try {
+            return factory.create();
```
honnix: Maybe `Objects.requireNonNull` to fail early.

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
```diff
@@ -156,4 +175,81 @@ public void exceptionsExitNonZero() throws Exception {
...
+
+    assertThat(status.get(), is(0));
+  }
+
+  @Test
+  public void testValidTerminationHook() {
+    final HookResult hookResult = new HookResult();
+
+    doAnswer(invocation -> {
+      Integer exitCode = invocation.getArgumentAt(0, Integer.class);
```
honnix: You can use `ArgumentCaptor` to simplify this a bit. Just fyi.

NatashaL: 💲 good to know. I realized I can simplify it even more though, no need for captors either

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
```diff
@@ -156,4 +175,81 @@ public void exceptionsExitNonZero() throws Exception {
 
     assertThat(status.get(), is(1));
   }
+
+  @Test
+  public void handleInvalidTerminationHookFactories() throws Exception {
```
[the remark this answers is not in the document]
NatashaL: Added more comprehensive names

honnix:
👍

PR #70, opened 2018-05-09
Remove StyxExecutionId and related config

NatashaL:


honnix:
It's green so 👍 
We will need to add codecov config. That would be another PR.

PR #71, opened 2018-05-15
merge flo-contrib and flo

honnix:
* clean up pom.xml
* pass typesafe config to TerminationHookFactory
* add flo-styx with TerminationLogging
* upgrade libs
* get logback completely out of the picture

@NatashaL @bergman PTAL


pom.xml
```diff
@@ -80,51 +90,87 @@
...
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bigquery</artifactId>
+        <version>1.29.0</version>
+      </dependency>
+
+      <!-- version resolution -->
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>25.0-jre</version>
```
bergman: this sounds way too recent, i've had problems with everything after 23 (when they added the `-jre` and `-android`)

honnix: That's more or less what we have in Styx production. :P
What kind of problems do we have?

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/TerminationLogging.java
```diff
@@ -0,0 +1,72 @@
...
+    this.config = Objects.requireNonNull(config);
+  }
+
+  @Override
+  public void accept(Integer exitCode) {
+    final String content = String.format("{\"component_id\": \"%s\","
+                                         + "\"workflow_id\": \"%s\","
+                                         + "\"parameter\": \"%s\","
+                                         + "\"execution_id\": \"%s\","
+                                         + "\"event\": \"terminated\","
```
NatashaL: I would assume we want `exited` here, not sure if it makes any difference for Styx. With edgedocker-base we log `terminated` (given SIGTERM), `interrupted` (given SIGINT) and `exited` (for natural termination).

honnix: right, that sounds better.

contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/BigQueryContext.java
```diff
@@ -135,4 +136,20 @@ TableId publish(StagingTableId stagingTableId) {
 
     return tableId;
   }
+
+  private static void checkJobStatus(Job job, TableId tableId) {
+    if (job != null && job.getStatus().getError() == null) {
```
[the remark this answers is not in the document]
honnix: By reading the code, I think it is still possible no exception thrown but the job failed. This API designing and doc are so confusing.

pom.xml
honnix: downgraded to 23.0

NatashaL:
@honnix the error handling seems like it covers all cases. Both if there's an exception thrown and if the error is contained. Don't see what more we can do about it for now. 

I think this is good for now. Any more comments or 👍  ? cc @bergman

PR #76, opened 2018-05-22
Flo extras

honnix:


contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/LabelUtil.java
```diff
@@ -0,0 +1,129 @@
...
+    if (!Character.isAlphabetic(value.charAt(0))) {
+      value = String.format("%s-%s", pad, value);
+    }
+    // Last character must be [a-z0-9] (not -)
+    if (value.charAt(value.length() - 1) == '-') {
+      value = String.format("%s%s", value, pad);
+    }
+    // Trim string if it is too long
+    if (value.length() > LABEL_VALUE_MAX_LENGTH) {
+      value = value.substring(0, LABEL_VALUE_MAX_LENGTH);
```
NatashaL: Seems like this can cause `value` to end up being an invalid string even after the applied pad suffix. Why don't we check and cut the length of the label before applying the pad? We can check for a shorter length - seems like 58 would be a good length - that would allow for the prefix and suffix (ex `tt-`... and ..`tt` to later be applied and the label to still be within the limit of 63.

honnix: Sure we can do that. Just copied what we did in internal Luigi code.

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/LabelUtil.java
```diff
@@ -0,0 +1,133 @@
...
+      value = value.substring(0, LABEL_VALUE_MAX_LENGTH - pad.length());
+    }
+
+    // Last character must be [a-z0-9] (not -)
+    if (value.charAt(value.length() - 1) == '-') {
+      value = String.format("%s%s", value, pad);
+    }
+
+    // Final check
+    if (!value.matches(LABEL_VALUE_REGEX)) {
```
honnix: This is pretty much being paranoid I think. Can't find a case.

NatashaL: Seems like it, can't think of a case for it either

contrib/flo-extras/src/main/java/com/spotify/flo/contrib/dataflow/PipelineOptionsSupplier.java
```diff
@@ -0,0 +1,108 @@
...
+
+  Optional<String> zone();
+
+  Optional<String> stagingLocation();
+
+  Optional<String> tempLocation();
+
+  Optional<String> gcpTempLocation();
+
+  Optional<AutoscalingAlgorithmType> autoscalingAlgorithm();
```
NatashaL: Shall we remove the `Optional` out of the fields that have default values? `maxNumWorkers` `network` `autoscalingAlgorithm` and `runner`. I initially made all optional for no specific reason. I think this should be revisited.

NatashaL: Some others might need to be required too (even without having a default value) such as `serviceAccount`?

honnix: Actually everything is optional except `project` in this case. Users can bypass `.builder()` and start from an empty builder and in that case we should perhaps still make them optional.

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/Constants.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.flo.contrib.styx;
+
+final class Constants {
+
+  Constants() {
```
bergman: just make it private?

contrib/flo-extras/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+  </description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-google-cloud-dataflow-java</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.norberg</groupId>
+      <artifactId>auto-matter</artifactId>
```
bergman: this can be for compile scope only, to not include it in flo jars

contrib/flo-extras/src/main/java/com/spotify/flo/contrib/dataflow/PipelineOptionsSupplier.java
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * The project ID for your Google Cloud Project.
+   * Mandatory value. Not setting it will result {@link NullPointerException}.
+   */
+  String project();
+
+  /**
+   * Run the job as a specific service account.
+   */
+  String serviceAccount();
```
bergman: also mandatory since not `Optional`

contrib/flo-styx/pom.xml
```diff
@@ -35,5 +39,13 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-google-cloud-dataflow-java</artifactId>
```
bergman: I feel this make flo-styx quite opinionated/biased towards dataflow. Is it really necessary?

contrib/flo-extras/pom.xml
honnix: it is `provided` declared in parent pom.

bergman:
As mentioned in line comment, I feel this is unnecessarily opinionated towards dataflow. I don't see how `flo-styx` would need to know about dataflow. How come `flo-extras` means `flo-dataflow` in practice?

honnix:
I actually think we have been fooling ourselves not opinionated towards dataflow.

honnix:
`labels` only works for `DataflowRunner`, so we will need this dependency anyway. We can perhaps returning `PipelineOptions` although to me that seems like a lier.

bergman:
I think we are painting ourselves into a corner if we lock our tools to a specific vendor. I'm all for having `flo-dataflow` with utils that make it easier to use dataflow etc.

Does it make sense for a flo dataproc job in styx to have a dependency on beam SDK?

contrib/flo-styx/pom.xml
bergman: There aren't many things in styx that are about data really. It works great for data but can be used for whatever you like. The same applies even more to flo. There is nothing concerning data in flo. Having separate modules that add functionality is fine but steering it too much into a specific use-case (data, styx, dataflow …) will not be good in the long run I think.

honnix:
They are part of flo contrib. I don't see why it is not suitable. What if we rename `flo-extras` to `flo-dataflow` and `flo-styx` to `flo-styx-dataflow`?

bergman:
It was beam altogether I was talking about, not dataflow vs beam. Sorry, could have been more clear on that.

I don't want you to move stuff into internal, I just don't want things to be too opinionated at this low level.

bergman:
Btw, how do service accounts and projects relate to running a beam job with the local runner?

PR #82, opened 2018-06-01
grpc context propagation with task name

danielnorberg:
Make outer grpc context and task name available to logging etc in user code (process fn).

### DIFF
--- a/flo-workflow/src/main/java/com/spotify/flo/Tracing.java
+++ b/flo-workflow/src/main/java/com/spotify/flo/Tracing.java
@@ -2,10 +2,13 @@ package com.spotify.flo;
 
 import io.grpc.Context;
 
-public class Tracing {
+public final class Tracing {
 
   public static final Context.Key<String> TASK_ID = Context.keyWithDefault("task-id", "");
   public static final Context.Key<String> TASK_NAME = Context.keyWithDefault("task-name", "");
   public static final Context.Key<String> TASK_ARGS = Context.keyWithDefault("task-args", "");
 
+  private Tracing() {
+    throw new UnsupportedOperationException();
+  }
 }
--- a/flo-workflow/src/main/java/com/spotify/flo/Tracing.java
+++ b/flo-workflow/src/main/java/com/spotify/flo/Tracing.java
@@ -4,6 +4,8 @@ import io.grpc.Context;
 
 public class Tracing {
 
+  public static final Context.Key<String> TASK_ID = Context.keyWithDefault("task-id", "");
   public static final Context.Key<String> TASK_NAME = Context.keyWithDefault("task-name", "");
+  public static final Context.Key<String> TASK_ARGS = Context.keyWithDefault("task-args", "");
 
 }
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -109,6 +109,11 @@
         <artifactId>kryo-shaded</artifactId>
         <version>4.0.0</version>
       </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-context</artifactId>
+        <version>1.12.0</version>
```
honnix: Internally we use 1.4.0 due to compatibility issue. If 1.4.0 can work in this case, we might use that.

flo-workflow/src/main/java/com/spotify/flo/context/TracingContext.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public <T> Value<T> invokeProcessFn(TaskId taskId, Fn<Value<T>> processFn) {
+    try {
+      return Context.current()
+          .withValue(TASK_ID, taskId.toString())
+          .withValue(TASK_NAME, taskId.name())
+          .withValue(TASK_ARGS, taskId.args())
+          .call(() -> super.invokeProcessFn(taskId, processFn));
+    } catch (Exception e) {
```
honnix: Will it look better having two catches?

flo-workflow/src/main/java/com/spotify/flo/Tracing.java
```diff
@@ -0,0 +1,11 @@
+package com.spotify.flo;
+
+import io.grpc.Context;
+
+public class Tracing {
```
honnix: This class needs a bit more touch, static, final, private constructor, etc.

flo-workflow/src/main/java/com/spotify/flo/context/TracingContext.java
```diff
@@ -0,0 +1,56 @@
...
+import static com.spotify.flo.Tracing.TASK_ID;
+
+import com.spotify.flo.EvalContext;
+import com.spotify.flo.Fn;
+import com.spotify.flo.Task;
+import com.spotify.flo.TaskId;
+import io.grpc.Context;
+
+/**
+ * A {@link EvalContext} that sets the current {@link Task#id()} on the {@link Context} when calling the process fn.
```
NatashaL: Update description

NatashaL:
👍

flo-workflow/src/main/java/com/spotify/flo/context/TracingContext.java
danielnorberg: How does it need to be updated?

NatashaL: Oh my bad, I read EvalContext != TracingContext and assumed it was a copy-paster

honnix: I would probably do the same. xD

PR #84, opened 2018-06-07
run task process fn in subprocess

danielnorberg:
By running the task in a subprocess we can capture all the output on std{out,err} and decorate it with task id etc.

Note: I do worry about the additional overhead in terms of memory usage for each task subprocess. If several tasks run concurrently, we might e.g. hit the memory limit of the container we're running in. The `reference.conf` currently specifies `flo.workers = 4` so we should see up to 5 JVMs running at one time.

### DIFF
--- a/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
+++ b/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
@@ -31,6 +31,7 @@ interface StructuredLogMessage {
   @JsonProperty @Nullable String logger();
   @JsonProperty @Nullable String thread();
   @JsonProperty @Nullable String message();
+  @JsonProperty @Nullable String framework();
   @JsonProperty @Nullable String styx_component_id();
   @JsonProperty @Nullable String styx_workflow_id();
   @JsonProperty @Nullable String styx_docker_args();
@@ -40,13 +41,13 @@ interface StructuredLogMessage {
   @JsonProperty @Nullable String styx_execution_id();
   @JsonProperty @Nullable String styx_trigger_id();
   @JsonProperty @Nullable String styx_trigger_type();
-  @JsonProperty @Nullable String flo_task_id();
-  @JsonProperty @Nullable String flo_task_name();
-  @JsonProperty @Nullable String flo_task_args();
+  @JsonProperty @Nullable String task_id();
+  @JsonProperty @Nullable String task_name();
+  @JsonProperty @Nullable String task_args();
 
   StructuredLogMessageBuilder builder();
 
   static StructuredLogMessageBuilder newBuilder() {
-    return new StructuredLogMessageBuilder();
+    return new StructuredLogMessageBuilder().framework("flo");
   }
 }
--- a/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
+++ b/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
@@ -1,0 +1,52 @@
+/*-
+ * -\-\-
+ * Flo Styx
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.flo.contrib.styx;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.norberg.automatter.AutoMatter;
+import javax.annotation.Nullable;
+
+@AutoMatter
+interface StructuredLogMessage {
+  @JsonProperty @Nullable String time();
+  @JsonProperty @Nullable String severity();
+  @JsonProperty @Nullable String logger();
+  @JsonProperty @Nullable String thread();
+  @JsonProperty @Nullable String message();
+  @JsonProperty @Nullable String styx_component_id();
+  @JsonProperty @Nullable String styx_workflow_id();
+  @JsonProperty @Nullable String styx_docker_args();
+  @JsonProperty @Nullable String styx_docker_image();
+  @JsonProperty @Nullable String styx_commit_sha();
+  @JsonProperty @Nullable String styx_parameter();
+  @JsonProperty @Nullable String styx_execution_id();
+  @JsonProperty @Nullable String styx_trigger_id();
+  @JsonProperty @Nullable String styx_trigger_type();
+  @JsonProperty @Nullable String flo_task_id();
+  @JsonProperty @Nullable String flo_task_name();
+  @JsonProperty @Nullable String flo_task_args();
+
+  StructuredLogMessageBuilder builder();
+
+  static StructuredLogMessageBuilder newBuilder() {
+    return new StructuredLogMessageBuilder();
+  }
+}
--- a/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
+++ b/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
@@ -22,32 +22,56 @@ package com.spotify.flo.contrib.styx;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.norberg.automatter.AutoMatter;
-import javax.annotation.Nullable;
 
 @AutoMatter
 interface StructuredLogMessage {
-  @JsonProperty @Nullable String time();
-  @JsonProperty @Nullable String severity();
-  @JsonProperty @Nullable String logger();
-  @JsonProperty @Nullable String thread();
-  @JsonProperty @Nullable String message();
-  @JsonProperty @Nullable String framework();
-  @JsonProperty @Nullable String styx_component_id();
-  @JsonProperty @Nullable String styx_workflow_id();
-  @JsonProperty @Nullable String styx_docker_args();
-  @JsonProperty @Nullable String styx_docker_image();
-  @JsonProperty @Nullable String styx_commit_sha();
-  @JsonProperty @Nullable String styx_parameter();
-  @JsonProperty @Nullable String styx_execution_id();
-  @JsonProperty @Nullable String styx_trigger_id();
-  @JsonProperty @Nullable String styx_trigger_type();
-  @JsonProperty @Nullable String task_id();
-  @JsonProperty @Nullable String task_name();
-  @JsonProperty @Nullable String task_args();
+  @AutoMatter
+  interface Styx {
+    @JsonProperty String component_id();
+    @JsonProperty String workflow_id();
+    @JsonProperty String docker_args();
+    @JsonProperty String docker_image();
+    @JsonProperty String commit_sha();
+    @JsonProperty String parameter();
+    @JsonProperty String execution_id();
+    @JsonProperty String trigger_id();
+    @JsonProperty String trigger_type();
 
-  StructuredLogMessageBuilder builder();
+    static StyxBuilder newBuilder() {
+      return new StyxBuilder();
+    }
+  }
+
+  @AutoMatter
+  interface Task {
+    @JsonProperty String id();
+    @JsonProperty String name();
+    @JsonProperty String args();
+
+    static TaskBuilder newBuilder() {
+      return new TaskBuilder();
+    }
+  }
+
+  @AutoMatter
+  interface Workflow {
+    @JsonProperty Styx styx();
+    @JsonProperty String framework();
+    @JsonProperty Task task();
+
+    static WorkflowBuilder newBuilder() {
+      return new WorkflowBuilder().framework("flo");
+    }
+  }
+
+  @JsonProperty String time();
+  @JsonProperty String severity();
+  @JsonProperty String logger();
+  @JsonProperty String thread();
+  @JsonProperty String message();
+  @JsonProperty Workflow workflow();
 
   static StructuredLogMessageBuilder newBuilder() {
-    return new StructuredLogMessageBuilder().framework("flo");
+    return new StructuredLogMessageBuilder();
   }
 }
--- a/flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
+++ b/flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
@@ -158,11 +158,12 @@ public final class FloRunner<T> {
 
     return
         TracingContext.composeWith(
-            MemoizingContext.composeWith(
-                OverridingContext.composeWith(
-                    LoggingContext.composeWith(
-                        baseContext, logging),
-                    logging)));
+            ForkingContext.composeWith(
+                MemoizingContext.composeWith(
+                    OverridingContext.composeWith(
+                        LoggingContext.composeWith(
+                            baseContext, logging),
+                        logging))));
   }
 
   private EvalContext createRootContext() {
--- a/flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
+++ b/flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
@@ -1,0 +1,149 @@
+/*-
+ * -\-\-
+ * Flo Runner
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.flo.context;
+
+import com.spotify.flo.EvalContext;
+import com.spotify.flo.Fn;
+import com.spotify.flo.Task;
+import com.spotify.flo.TaskId;
+import com.spotify.flo.Tracing;
+import io.grpc.Context;
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An {@link EvalContext} that runs tasks in sub-processes.
+ * <p>
+ * Forking can be disabled using the environment variable {@code FLO_DISABLE_FORKING=true}.
+ * <p>
+ * Forking is disabled by default when running in the debugger, but can be enabled by {@code FLO_FORCE_FORK=true}.
+ * <p>
+ * The basic idea here is to intercept the process fn by overriding {@link EvalContext#invokeProcessFn(TaskId, Fn)},
+ * and executing it in a sub-process JVM. The process fn closure and the result is transported in to and out of the
+ * sub-process using serialization.
+ */
+class ForkingContext implements EvalContext {
+
+  private static final Logger log = LoggerFactory.getLogger(ForkingContext.class);
+
+  // This is marked transient so as to avoid serializing the whole stack of EvalContexts
+  private transient final EvalContext delegate;
+
+  private ForkingContext(EvalContext delegate) {
+    this.delegate = Objects.requireNonNull(delegate);
+  }
+
+  static EvalContext composeWith(EvalContext baseContext) {
+    // Is the Java Debug Wire Protocol activated?
+    final boolean inDebugger = ManagementFactory.getRuntimeMXBean().
+        getInputArguments().stream().anyMatch(s -> s.contains("-agentlib:jdwp"));
+
+    final Optional<Boolean> forking = Optional.ofNullable(System.getenv("FLO_FORKING")).map(Boolean::parseBoolean);
+
+    if (forking.isPresent()) {
+      if (forking.get()) {
+        log.debug("Forking enabled (environment variable FLO_FORKING=true)");
+        return new ForkingContext(baseContext);
+      } else {
+        log.debug("Forking disabled (environment variable FLO_FORKING=true)");
+        return baseContext;
+      }
+    } else if (inDebugger) {
+      log.debug("Debugger detected, forking disabled by default "
+          + "(enable by setting environment variable FLO_FORKING=true)");
+      return baseContext;
+    } else {
+      log.debug("Debugger not detected, forking enabled by default "
+          + "(disable by setting environment variable FLO_FORKING=false)");
+      return new ForkingContext(baseContext);
+    }
+  }
+
+  @Override
+  public <T> Value<T> evaluateInternal(Task<T> task, EvalContext context) {
+    return delegate().evaluateInternal(task, context);
+  }
+
+  @Override
+  public <T> Value<T> value(Fn<T> value) {
+    // Note: This method is called from within the process fn lambda, and thus this ForkingContext instance will be
+    // captured in the closure. This method will then be called in the task sub-process. As the delegate field is
+    // transient it will be null and delegate() will return a SyncContext that will immediately call the value fn.
+    return delegate().value(value);
+  }
+
+  @Override
+  public <T> Value<T> immediateValue(T value) {
+    return delegate().immediateValue(value);
+  }
+
+  @Override
+  public <T> Promise<T> promise() {
+    return delegate().promise();
+  }
+
+  @Override
+  public <T> Value<T> invokeProcessFn(TaskId taskId, Fn<Value<T>> processFn) {
+    if (delegate == null) {
+      throw new UnsupportedOperationException("nested execution not supported");
+    }
+    // Wrap the process fn in a lambda that will execute the original process fn closure in a sub-process.
+    final Fn<Value<T>> forkingProcessFn = fork(taskId, processFn);
+    // Pass on the wrapped process fn to let the rest of EvalContexts do their thing. The last EvalContext in the chain
+    // will invoke the wrapped lambda, causing the sub-process execution to happen there.
+    return delegate.invokeProcessFn(taskId, forkingProcessFn);
+  }
+
+  private EvalContext delegate() {
+    // The delegate field will be null if this is in the sub-process
+    if (this.delegate == null) {
+      return SyncContext.create();
+    } else {
+      return delegate;
+    }
+  }
+
+  private <T> Fn<Value<T>> fork(TaskId taskId, Fn<Value<T>> value) {
+    return () -> {
+      try (final ForkingExecutor executor = new ForkingExecutor()) {
+        executor.environment(Collections.singletonMap("FLO_TASK_ID", taskId.toString()));
+        return immediateValue(executor.execute(() -> {
+          try {
+            return Context.current()
+                .withValue(Tracing.TASK_ID, taskId)
+                .call(value::get);
+          } catch (RuntimeException e) {
+            throw e;
+          } catch (Exception e) {
+            throw new RuntimeException(e);
+          }
+        }));
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    };
+  }
+}
--- a/flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
+++ b/flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
@@ -22,14 +22,12 @@ package com.spotify.flo.context;
 
 import com.spotify.flo.EvalContext;
 import com.spotify.flo.Fn;
-import com.spotify.flo.Task;
 import com.spotify.flo.TaskId;
 import com.spotify.flo.Tracing;
 import io.grpc.Context;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.util.Collections;
-import java.util.Objects;
 import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -45,15 +43,12 @@ import org.slf4j.LoggerFactory;
  * and executing it in a sub-process JVM. The process fn closure and the result is transported in to and out of the
  * sub-process using serialization.
  */
-class ForkingContext implements EvalContext {
+class ForkingContext extends ForwardingEvalContext {
 
   private static final Logger log = LoggerFactory.getLogger(ForkingContext.class);
 
-  // This is marked transient so as to avoid serializing the whole stack of EvalContexts
-  private transient final EvalContext delegate;
-
   private ForkingContext(EvalContext delegate) {
-    this.delegate = Objects.requireNonNull(delegate);
+    super(delegate);
   }
 
   static EvalContext composeWith(EvalContext baseContext) {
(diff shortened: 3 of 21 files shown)


honnix:
omg this is hairy. :(

honnix:
awesome hack indeed!

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -178,7 +178,7 @@ private EvalContext createRootContext() {
           config.getInt("flo.workers"),
           threadFactory);
       closeables.add(executorCloser(executor));
-      return EvalContext.async(executor);
+      return ForkingEvalContext.composeWith(EvalContext.async(executor));
```
danielnorberg: Add a check here to not use the `ForkingEvalContext` if running locally

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,319 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ForkingEvalContext extends ForwardingEvalContext {
+
+  private static final Logger log = LoggerFactory.getLogger(ForwardingEvalContext.class);
+
+  public ForkingEvalContext(EvalContext delegate) {
```
honnix: This can be private.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,319 @@
...
+      Watchdog() {
+        setDaemon(false);
+      }
+
+      @Override
+      public void run() {
+        // Wait for parent to exit.
+        try {
+          while (true) {
+            int c = System.in.read();
```
honnix: If parent process exits, here we will get EOF?

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,319 @@
...
+          return result;
+        }
+      } finally {
+        executor.shutdownNow();
+        tryDeleteDir(tempdir);
+      }
+    };
+  }
+
+  private void tryDeleteDir(Path path) {
```
honnix: amazing still so much code needed to recursively delete a dir in Java.

honnix:
Overall looks good.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: 👍

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: Yeah

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: My favorite is shelling out to `rm -rf`. Battle tested and hard to beat!

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,344 @@
...
+        // Outsource the problem completely to the logger implementation provided by the user:
+        // 1. No special setup for the child process, it just logs to std{out,err} as configured by user.
+        // 2. Parent process tails process std{out,err} and re-logs each line using the slf4j logger. Any structured
+        //    log messages now end up wrapped in the message field of the log message.
+        // 3. The parent process logger inspects all log messages and bypasses normal processing for any (wrapped)
+        //    structured log messages, just decorating it with task metadata and emitting it as is.
+
+        // Note:
+        // * Child process does not need grpc context
+        // * Tempted at this point to run the child task in a container and let GKE deal with the logs
```
honnix: A scheduler inside a scheduler. :D
We can as well abandon orchestration and let Styx handle it.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,344 @@
...
+        // 2. Make the parent tail the log file, parse the json messages, decorate with task metadata and re-emit the
+        //    log messages. Note that re-emitting the log messages through the parent process slf4j logger would involve
+        //    figuring out how to keep the original log message metadata (e.g. timestamp, logger, thread, etc). Maybe
+        //    using the LocationAwareLogger interface. Printing the log messages directly to stderr might be more
+        //    straightforward. flo does not pull in a logging implementation, we'd have to pull one in (e.g. logback)
+        //    and set it up to do structured logging in the child process. Alternatively roll our own.
+        // 3. Also tail child process std{out,err} for unstructured output and log each line using the parent process
+        //    logger, taking care to not choke on huge lines and non-text output.
+
+        // Plan B
```
honnix: This seems less disruptive.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: \o/

danielnorberg: Make flo just emit the dag and let styx run the tasks?

honnix:
Would this lib help a bit? https://github.com/zeroturnaround/zt-exec/blob/master/README.md

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: Went with [Plan C](https://github.com/spotify/flo/pull/84/commits/0534809920c445a725428fe6d9275220e88e6fc8#diff-171892128c20990cf4ce11dd33d5ca06R140):

```
Keep it simple.
1. Add FLO_TASK_ID to child process environment.
2. Rely on user provided logger to pick up FLO_TASK_ID.
3. Parent process just copies std{err,out} line by line. The reason for manual line-by-line copying instead
   of using Redirect.INHERIT is to ensure that line contents are not interleaved.
```

rouzwawi:
Is `Value<T> value(Fn<T> value)` the right place to fork? IIRC `Value<T>`s are created in more places than just for new tasks, which would mean a lot of forking for things that are just internal, immediate values.

I think forking from `invokeProcessFn` would be more precise.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
danielnorberg: Done in `ForkingEvalContext` itself.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -178,7 +178,7 @@ private EvalContext createRootContext() {
...
     } else {
       return EvalContext.sync();
```
danielnorberg: Also wrap this context

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,382 @@
...
+            throw (RuntimeException) error;
+          } else {
+            throw new RuntimeException(error);
+          }
+        } else {
+          // Success
+          log.debug("subprocess exited with result file");
+          final T result;
+          try {
+            result = PersistingContext.deserialize(resultFile);
```
honnix: Can we return from here already? Or maybe it doesn't compile due to some rules?

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,382 @@
...
+import java.util.concurrent.Executors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An {@link EvalContext} that runs tasks in sub-processes.
+ * <p>
+ * Forking can be disabled using the environment variable {@code FLO_DISABLE_FORKING=true}.
+ * <p>
+ * Forking is disabled by default when running in the debugger, but can be enabled by {@code FLO_FORCE_FORK=true}.
```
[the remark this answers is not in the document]
danielnorberg: 👍

danielnorberg:
@rouzwawi forking in `Value<T> value(Fn<T> value)` seemed to be the way to serialize the process fn closure without any additional flo frames in the call stack. The `invokeProcessFn` method results in a call to `value()`, traversing all the `EvalContext` layers before ending up in `AsyncContext.value()` where the evaluation is executed on the thread pool.

See this (async) stack trace captured with intellij:

```
lambda$taskIdIsInContext$648bafdb$1:248, FloRunnerTest (com.spotify.flo.context)
get:-1, 686466458 (com.spotify.flo.context.FloRunnerTest$$Lambda$13)
get:-1, 1489069835 (com.spotify.flo.TaskBuilderImpl$Builder0$$Lambda$8)
run$$$capture:1590, CompletableFuture$AsyncSupply (java.util.concurrent)
 - Async stacktrace
supplyAsync:-1, CompletableFuture (java.util.concurrent)
value:61, AsyncContext (com.spotify.flo.context)
value:52, ForwardingEvalContext (com.spotify.flo.context)
value:69, LoggingContext (com.spotify.flo.context)
value:52, ForwardingEvalContext (com.spotify.flo.context)
value:52, ForwardingEvalContext (com.spotify.flo.context)
value:52, ForwardingEvalContext (com.spotify.flo.context)
value:52, ForwardingEvalContext (com.spotify.flo.context)
lambda$null$1cd2cf8d$1:44, BuilderUtils (com.spotify.flo)
lambda$invokeProcessFn$e9e1fce5$1:218, MemoizingContext (com.spotify.flo.context)
invokeProcessFn:90, EvalContext (com.spotify.flo)
invokeProcessFn:122, InstrumentedContext (com.spotify.flo.context)
invokeProcessFn:57, LoggingContext (com.spotify.flo.context)
invokeProcessFn:47, ForwardingEvalContext (com.spotify.flo.context)
invokeProcessFn:218, MemoizingContext (com.spotify.flo.context)
invokeProcessFn:47, ForwardingEvalContext (com.spotify.flo.context)
lambda$invokeProcessFn$0:49, TracingContext (com.spotify.flo.context)
call:577, Context (io.grpc)
invokeProcessFn:49, TracingContext (com.spotify.flo.context)
invokeProcessFn:47, ForwardingEvalContext (com.spotify.flo.context)
lambda$gated$8ebcb94d$1:44, BuilderUtils (com.spotify.flo)
evaluateInternal:71, EvalContext (com.spotify.flo)
lambda$evaluateInternal$3e9ca674$1:56, AsyncContext (com.spotify.flo.context)
run$$$capture:1590, CompletableFuture$AsyncSupply (java.util.concurrent)
run:-1, CompletableFuture$AsyncSupply (java.util.concurrent)
 - Async stacktrace
supplyAsync:-1, CompletableFuture (java.util.concurrent)
flatten:75, AsyncContext (com.spotify.flo.context)
evaluateInternal:56, AsyncContext (com.spotify.flo.context)
evaluateInternal:115, InstrumentedContext (com.spotify.flo.context)
evaluateInternal:50, LoggingContext (com.spotify.flo.context)
evaluateInternal:71, OverridingContext (com.spotify.flo.context)
evaluate:310, MemoizingContext$EvalBundle (com.spotify.flo.context)
evaluateInternal:199, MemoizingContext (com.spotify.flo.context)
evaluateInternal:42, ForwardingEvalContext (com.spotify.flo.context)
evaluate:55, EvalContext (com.spotify.flo)
run:125, FloRunner (com.spotify.flo.context)
runTask:89, FloRunner (com.spotify.flo.context)
runTask:110, FloRunner (com.spotify.flo.context)
```

rouzwawi:
@danielnorberg Refreshing myself a bit on the mechanics.

I'm thinking that having `value(...)` calls fork a new JVM might add too much overhead for things that are just intermediate values and not the actual task `process` fn (e.g. when `evalContext.immediateValue(x)` is called). I think this is done at least once for each task input (see implementation of `_context()` and `inputs()` when using `toValueList(ec)`).

`invokeProcessFn` returning a `Value<T>` that might not be ready should not be a problem as long as you have the forked JVM wait for that value to complete before exiting and completing the parent process value.

If the forking context is at the end of the chain, the only meaningful frames that would be dropped in the forked process would be `value:69, LoggingContext (com.spotify.flo.context)` which seems benign as we're dealing with logging differently here anyways.

If the `EvalContext` used in the forked process runs a bare bone `SyncContext`, the value call made from the returned `Value<T>` of `invokeProcessFn` would be sync and blocking:

```
(a, b) -> ec.invokeProcessFn(taskId, () -> ec.value(() -> code.apply(a, b)))
```

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,382 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(ForwardingEvalContext.class);
+
+  // Is the Java Debug Wire Protocol activated?
+  private static boolean IN_DEBUGGER = ManagementFactory.getRuntimeMXBean().
+      getInputArguments().stream().anyMatch(s -> s.contains("-agentlib:jdwp"));
+
+  private static boolean FLO_DISABLE_FORKING = Boolean.parseBoolean(System.getenv("FLO_DISABLE_FORKING"));
+
+  private static boolean FORCE_FORK = Boolean.parseBoolean(System.getenv("FLO_FORCE_FORK"));
```
rouzwawi: Wouldn't it be cleaner to just deal with this in `FloRunner` and skip adding `ForkingEvalContext` in the chain? That would more clearly separate concerns.

rouzwawi:
err. ignore the last part about creating a `SyncContext` in the subprocess, I was thinking about how I did it in `TaskRunnerEntrypoint` in my freezer prototype. This will only run an `Fn<T>` so no need for a context.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
honnix: How can we avoid adding `ForkingEvalContext` in the chain? I'm not sure I understand this.

rouzwawi: In https://github.com/spotify/flo/blob/master/flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java#L153, there's some logic already that skips adding the PersistingContext, or selecting the base sync/async context. Using the ForkingEvalContext could be added (or not) in the same way).

honnix:
After reading `invokeProcessFn` I also got confused now. :/

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: I'll take a stab at it, perhaps by putting the check in `composeWith` or a `maybeComposeWith`. Or would it make more sense to put the environment variable lookups and decisions etc in FloRunner?

danielnorberg:
@rouzwawi `AsyncContext.immediateValue()` returns a completed future directly without calling `value()`, so I guess `immediateValue()` shouldn't fork. The same is not true for `SyncContext` though as it does not explicitly implement `immediateValue()`. Perhaps `EvalContext.immediateValue()` should not have a default implementation that calls `value()`?

Still, it would be good to somehow precisely be able to only fork for the execution of the process fn and then also only execute the process fn in the child process, correctly invoking the delegate EC in the main process.

> invokeProcessFn returning a Value<T> that might not be ready should not be a problem as long as you have the forked JVM wait for that value to complete before exiting and completing the parent process value.

I'm not sure if I entirely follow...

My intention (as things are implemented now) was run and wait for the child process to complete in a blocking fashion on the `AsyncContext` thread pool. Hence `ForkingEvalContext.value()` calling [`super.value()`](https://github.com/spotify/flo/pull/84/files#diff-171892128c20990cf4ce11dd33d5ca06R83) with a `Fn<T> value` that's been wrapped in a `Fn<T>` that forks. As a bonus, there's no disruption to functioning of the rest of the EvalContexts in the stack. It's almost as if the user themselves had implemented the fork in their process fn, like:

```java
Task.named("foo").ofType(String.class)
  .process(() -> {
    fork(() -> {
    	// do some work
    })
  });
```

Or even:
```java
Task.named("foo").ofType(String.class)
  .process(forking(() -> {
  	// do some work
  }));
```

Maybe that's what I'm missing here, the place where I can neatly wrap only the process fn in this forking thing.

Maybe that can be achieved using `invokeProcessFn`, but it's still not immediately clear to me. I'll bang my head against it some more. Worst case I guess we can invent some specialized method to do this e.g. `decorateProcessFn` etc.

Btw, it is still kind of puzzling to me that `invokeProcessFn` doesn't actually invoke the process fn. Is that intentional? The docs seem to indicate otherwise? https://github.com/spotify/flo/blob/master/flo-workflow/src/main/java/com/spotify/flo/EvalContext.java#L82 

It says
```java
@return The value of the process function invocation
```
But it actually only returns the handle?

honnix:
We talked about having `taskValue` to wrap user supplied function. That would require some changes in TaskBuilder I think.

honnix:
Check our internal ticket. I pasted the conversation there.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,385 @@
...
+      if (args.length != 3) {
+        err("args.length != 3");
+        System.exit(1);
+        return;
+      }
+      final Path closureFile = Paths.get(args[0]);
+      final Path resultFile = Paths.get(args[1]);
+      final Path errorFile = Paths.get(args[2]);
+
+      Context.current().withValue(Tracing.TASK_ID, taskId).run(() ->
```
honnix: How does this solve other thread than main outputting logs?

honnix: Or the encoder will check both environment and context?

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,385 @@
...
+        // Trampoline arguments
+        processBuilder.command().add(Trampoline.class.getName());
+        processBuilder.command().add(closureFile.toString());
+        processBuilder.command().add(resultFile.toString());
+        processBuilder.command().add(errorFile.toString());
+
+        // Propagate TASK_ID to child process
+        final TaskId taskId = Tracing.TASK_ID.get();
+        if (taskId != null) {
+          processBuilder.environment().put("FLO_TASK_ID", taskId.toString());
```
honnix: I hope Java by default carries over all environment to subprocess.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
```diff
@@ -0,0 +1,440 @@
...
+    if (delegate == null) {
+      throw new UnsupportedOperationException("nested execution not supported");
+    }
+    final Fn<Value<T>> forkingProcessFn = fork(processFn);
+    return delegate.invokeProcessFn(taskId, forkingProcessFn);
+  }
+
+  private EvalContext delegate() {
+    // In sub-process?
+    if (this.delegate == null) {
```
honnix: When does this happen?

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  @Override
+  public <T> Promise<T> promise() {
+    return delegate().promise();
+  }
+
+  @Override
+  public <T> Value<T> invokeProcessFn(TaskId taskId, Fn<Value<T>> processFn) {
+    if (delegate == null) {
```
honnix: When would this happen?

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
```diff
@@ -0,0 +1,440 @@
...
+      throw new UnsupportedOperationException("nested execution not supported");
+    }
+    final Fn<Value<T>> forkingProcessFn = fork(processFn);
+    return delegate.invokeProcessFn(taskId, forkingProcessFn);
+  }
+
+  private EvalContext delegate() {
+    // In sub-process?
+    if (this.delegate == null) {
```
honnix: How?

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: I guess I don't want to break the feature that the task-id is available in the task process fn, partly because it's already on master (we merged that PR and there's tests for it) and it might still come in handy.

flo-runner/src/main/java/com/spotify/flo/context/ForkingEvalContext.java
danielnorberg: It does.

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
danielnorberg: In the subprocess

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
danielnorberg: In the subprocess. It even says in the comment XD

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
honnix: if a user uses the `processWithContext` method.

danielnorberg: yeah

danielnorberg:
Current stack trace from `FloRunner.runTask` all the way to the task process fn.

```
com.spotify.flo.context.FloRunnerTest.lambda$taskIdIsInContext$648bafdb$1(FloRunnerTest.java:299)
com.spotify.flo.context.SyncContext.value(SyncContext.java:57)
com.spotify.flo.context.ForkingContext.value(ForkingContext.java:103)
com.spotify.flo.context.ForwardingEvalContext.value(ForwardingEvalContext.java:52)
com.spotify.flo.context.ForwardingEvalContext.value(ForwardingEvalContext.java:52)
com.spotify.flo.BuilderUtils.lambda$null$1cd2cf8d$1(BuilderUtils.java:44)
com.spotify.flo.context.ForkingContext$Trampoline.run(ForkingContext.java:362)
com.spotify.flo.context.ForkingContext$Trampoline.lambda$main$0(ForkingContext.java:344)
io.grpc.Context.run(Context.java:563)
com.spotify.flo.context.ForkingContext$Trampoline.main(ForkingContext.java:343)

- Fork

lambda$fork$cc93a42d$1:189, ForkingContext (com.spotify.flo.context)
get:-1, 304108971 (com.spotify.flo.context.ForkingContext$$Lambda$41)
lambda$invokeProcessFn$e9e1fce5$1:218, MemoizingContext (com.spotify.flo.context)
get:-1, 467775377 (com.spotify.flo.context.MemoizingContext$$Lambda$42)
invokeProcessFn:90, EvalContext (com.spotify.flo)
invokeProcessFn:122, InstrumentedContext (com.spotify.flo.context)
invokeProcessFn:57, LoggingContext (com.spotify.flo.context)
invokeProcessFn:47, ForwardingEvalContext (com.spotify.flo.context)
invokeProcessFn:218, MemoizingContext (com.spotify.flo.context)
invokeProcessFn:122, ForkingContext (com.spotify.flo.context)
invokeProcessFn:47, ForwardingEvalContext (com.spotify.flo.context)
lambda$invokeProcessFn$0:49, TracingContext (com.spotify.flo.context)
call:-1, 1249839702 (com.spotify.flo.context.TracingContext$$Lambda$40)
call:577, Context (io.grpc)
invokeProcessFn:49, TracingContext (com.spotify.flo.context)
invokeProcessFn:47, ForwardingEvalContext (com.spotify.flo.context)
lambda$gated$8ebcb94d$1:44, BuilderUtils (com.spotify.flo)
eval:-1, 371800738 (com.spotify.flo.BuilderUtils$$Lambda$9)
evaluateInternal:71, EvalContext (com.spotify.flo)
lambda$evaluateInternal$3e9ca674$1:56, AsyncContext (com.spotify.flo.context)
get:-1, 2031951755 (com.spotify.flo.context.AsyncContext$$Lambda$30)
run$$$capture:1590, CompletableFuture$AsyncSupply (java.util.concurrent)
 - Async stacktrace
supplyAsync:-1, CompletableFuture (java.util.concurrent)
flatten:75, AsyncContext (com.spotify.flo.context)
evaluateInternal:56, AsyncContext (com.spotify.flo.context)
evaluateInternal:115, InstrumentedContext (com.spotify.flo.context)
evaluateInternal:50, LoggingContext (com.spotify.flo.context)
evaluateInternal:71, OverridingContext (com.spotify.flo.context)
evaluate:310, MemoizingContext$EvalBundle (com.spotify.flo.context)
evaluateInternal:199, MemoizingContext (com.spotify.flo.context)
evaluateInternal:95, ForkingContext (com.spotify.flo.context)
evaluateInternal:42, ForwardingEvalContext (com.spotify.flo.context)
evaluate:55, EvalContext (com.spotify.flo)
run:125, FloRunner (com.spotify.flo.context)
runTask:89, FloRunner (com.spotify.flo.context)
runTask:110, FloRunner (com.spotify.flo.context)
...
```

flo-runner/src/main/java/com/spotify/flo/context/ForkingExecutor.java
```diff
@@ -0,0 +1,378 @@
...
+    }
+  }
+
+  private static void deleteDir(Path path) throws IOException {
+    try {
+      Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
+        @Override
+        public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+          try {
+            Files.delete(file);
```
honnix: What would cause concurrent deletion, or just to be extra cautious?

danielnorberg: `close()` is called both [here](https://github.com/spotify/flo/pull/84/files#diff-2df9957d7823bf437f9f3c51ceff75a6R87) and [here](https://github.com/spotify/flo/pull/84/files#diff-2df9957d7823bf437f9f3c51ceff75a6R82). A bit extra cautios code I guess.

honnix:
merged structure log PR, so anything from `flo-styx` has been reviewed.

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
```diff
@@ -0,0 +1,52 @@
...
+  @JsonProperty @Nullable String styx_component_id();
+  @JsonProperty @Nullable String styx_workflow_id();
+  @JsonProperty @Nullable String styx_docker_args();
+  @JsonProperty @Nullable String styx_docker_image();
+  @JsonProperty @Nullable String styx_commit_sha();
+  @JsonProperty @Nullable String styx_parameter();
+  @JsonProperty @Nullable String styx_execution_id();
+  @JsonProperty @Nullable String styx_trigger_id();
+  @JsonProperty @Nullable String styx_trigger_type();
+  @JsonProperty @Nullable String flo_task_id();
```
honnix: maybe we should remove all `flo_` prefix. that would make per task logging filter in stackdriver easier to compose.

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
```diff
@@ -0,0 +1,112 @@
...
+ * The basic idea here is to intercept the process fn by overriding {@link EvalContext#invokeProcessFn(TaskId, Fn)},
+ * and executing it in a sub-process JVM. The process fn closure and the result is transported in to and out of the
+ * sub-process using serialization.
+ */
+class ForkingContext extends ForwardingEvalContext {
+
+  private static final Logger log = LoggerFactory.getLogger(ForkingContext.class);
+
+  private ForkingContext(EvalContext delegate) {
+    super(delegate);
```
honnix: Where did `transient` go?

danielnorberg: The `EvalContext` is no longer part of the process fn closure, so it won't be serialized.

https://github.com/spotify/flo/pull/89/files#diff-7605014c7b48e15103f2b401b5f8ccd5L134

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
```diff
@@ -40,13 +41,13 @@
   @JsonProperty @Nullable String styx_execution_id();
   @JsonProperty @Nullable String styx_trigger_id();
   @JsonProperty @Nullable String styx_trigger_type();
-  @JsonProperty @Nullable String flo_task_id();
```
danielnorberg: Why removing the `flo_` prefix here?

honnix: to make it easier later when filtering luigi task logs
so we don’t need to differentiate
which could be tricky for UI work.

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
honnix: OK didn't realise that change. 👍

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
danielnorberg: Right. But `framework` and `task_id` feel a bit generic. Wondering if we should at least qualify them with `workflow_` or maybe nest them in a `"workflow": { ... }` object.

```yaml
"workflow": {
  "framework": "flo",
  "task_id": "my-task#54354",
  ...
}
```

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
```diff
@@ -0,0 +1,53 @@
...
+  @JsonProperty @Nullable String styx_docker_args();
+  @JsonProperty @Nullable String styx_docker_image();
+  @JsonProperty @Nullable String styx_commit_sha();
+  @JsonProperty @Nullable String styx_parameter();
+  @JsonProperty @Nullable String styx_execution_id();
+  @JsonProperty @Nullable String styx_trigger_id();
+  @JsonProperty @Nullable String styx_trigger_type();
+  @JsonProperty @Nullable String task_id();
+  @JsonProperty @Nullable String task_name();
+  @JsonProperty @Nullable String task_args();
```
ulzha: Wow, will this big message have repercussions in terms of log sizes?

danielnorberg: I think the logging volume from our workflows is small compared to the dataflow & dataproc log volume. Maybe we should gather metrics on this.

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
honnix: maybe just prefix with `workflow_`? no strong need to go nested i think.

honnix: those `styx_` things also fit in to `workflow` object if we want to go nested way.

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
honnix: Can we get that for free from stackdriver?

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
honnix: https://github.com/spotify/flo/pull/93

Zatine:
:+1:

danielnorberg:
yolo!

PR #85, opened 2018-06-11
run task in isolated class loader

danielnorberg:
An alternative to #84


danielnorberg:
@rouzwawi any thoughts on process vs class loader isolation?

danielnorberg:
Closing in favor of #84

PR #86, opened 2018-06-12
structured logging for styx

danielnorberg:
Continuation of #84. Can be repurposed on top of another branch/path forward if we choose to.

TODO:
- [x] Test reading task id from grpc context (in-process)

### DIFF
--- a/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLoggingEncoder.java
+++ b/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLoggingEncoder.java
@@ -105,18 +105,7 @@ public class StructuredLoggingEncoder extends EncoderBase<ILoggingEvent> {
   }
 
   private TaskId taskId() {
-    final TaskId taskId;
-    if (envTaskId != null) {
-      taskId = envTaskId;
-    } else {
-      final TaskId tracingTaskId = Tracing.TASK_ID.get();
-      if (tracingTaskId != null) {
-        taskId = tracingTaskId;
-      } else {
-        taskId = null;
-      }
-    }
-    return taskId;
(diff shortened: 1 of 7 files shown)


contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLoggingEncoder.java
```diff
@@ -0,0 +1,179 @@
...
+      WRITER.writeValue(baos, builder.build());
+      baos.write(LINE_SEPARATOR_BYTES);
+      return baos.toByteArray();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private byte[] encodeText(ILoggingEvent event) {
+    final TaskId taskId = Tracing.TASK_ID.get();
```
honnix: This needs to read from env as well.

danielnorberg: Good catch

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLoggingEncoder.java
```diff
@@ -0,0 +1,202 @@
...
+    try (final ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+      WRITER.writeValue(baos, builder.build());
+      baos.write(LINE_SEPARATOR_BYTES);
+      return baos.toByteArray();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private TaskId taskId() {
```
honnix: A ternary operator should be good enough I think.

honnix:
👍

PR #89, opened 2018-06-12
Remove processWithContext

rouzwawi:
Related to discussions around #84 

Removed the (mostly unused) `processWithContext` builder method, with the goal of removing the async call-frame indirection that was imposed on `EvalContext.invokeProcessFn`.

The result of this PR should be that the `Fn<T> processFn` passed into `EvalContext.invokeProcessFn` should actually be the user code, which should allow the context to decide how to invoke it without any `Value<T>` instances being created inside.

### DIFF
--- a/flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
+++ b/flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
@@ -91,9 +91,8 @@ public interface EvalContext {
   }
 
   /**
-   * When called from within any of the functions passed to {@code processWithContext}, this
-   * method will return the {@link Task} currently being processed. Otherwise an empty value
-   * will be returned.
+   * When called during the evaluation of a task, this method will return the {@link Task}
+   * currently being processed. Otherwise an empty value will be returned.
    *
    * <p>The return value of this method is stable for each instance of {@link EvalContext} that is
    * passed into the process functions. Calls from multiple threads will see the same result as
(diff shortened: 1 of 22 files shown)


rouzwawi:
@danielnorberg @honnix PTAL

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
```diff
@@ -79,21 +79,21 @@
...
-    return processFn.get();
+  default <T> Value<T> invokeProcessFn(TaskId taskId, Fn<T> processFn) {
+    return value(processFn);
   }
 
   /**
-   * When called from within any of the functions passed to {@code processWithContext}, this
-   * method will return the {@link Task} currently being processed. Otherwise an empty value
-   * will be returned.
+   * When called during the evaluation of a task, this method will return the {@link Task}
```
danielnorberg: If we remove `processWithContext` and thus user code has no way to get at the `EvalContext`, what would call this method? Does it make sense to keep it?

danielnorberg:
👍 Tried rebasing #84 on this and it works 👍

danielnorberg:
@rouzwawi ok to merge this so we can base our PRs on it?

rouzwawi:
Yes. We can look the taskId method in another PR.

danielnorberg:
\o/

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
rouzwawi: I noticed that `currentTask()` is actually used from the builder code as part of the `context()` implementation. So it has to stay.

PR #90, opened 2018-06-13
fix exception serialization

danielnorberg:
https://github.com/EsotericSoftware/kryo/issues/532

### DIFF
--- a/flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
+++ b/flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
@@ -1,0 +1,81 @@
+/*-
+ * -\-\-
+ * Flo Freezer
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.flo.freezer;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertThat;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.nio.file.Path;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class PersistingContextTest {
+
+  @Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test
+  public void exceptionSerialization() throws Exception {
+    final RuntimeException bar = new RuntimeException("bar");
+    bar.addSuppressed(new IOException("baz", new InterruptedException("quux")));
+    final Exception exception1 = new Exception("foo", bar);
+    exception1.addSuppressed(new FoobarException());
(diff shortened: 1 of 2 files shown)


flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
```diff
@@ -0,0 +1,81 @@
...
+    final Exception exception = exception1;
+
+    final Path exceptionFile = temporaryFolder.getRoot().toPath().resolve("exception");
+    PersistingContext.serialize(exception, exceptionFile);
+
+    final Exception deserialized = PersistingContext.deserialize(exceptionFile);
+
+    // Verify that stack trace can be accessed and printed and seems to be correct
+
+    deserialized.printStackTrace();
```
honnix: Is this needed?

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
```diff
@@ -0,0 +1,81 @@
...
+    assertThat(stacktrace, containsString("Suppressed: com.spotify.flo.freezer.PersistingContextTest$FoobarException"));
+    assertThat(stacktrace, containsString("Caused by: java.lang.RuntimeException: bar"));
+    assertThat(stacktrace, containsString("Suppressed: java.io.IOException: baz"));
+    assertThat(stacktrace, containsString("Caused by: java.lang.InterruptedException: quux"));
+
+    assertThat(deserialized.getStackTrace().length, is(not(0)));
+    assertThat(deserialized.getStackTrace()[0].getClassName(), is("com.spotify.flo.freezer.PersistingContextTest"));
+    assertThat(deserialized.getStackTrace()[0].getMethodName(), is("exceptionSerialization"));
+
+    for (StackTraceElement stackTraceElement : deserialized.getStackTrace()) {
```
honnix: Here as well.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
danielnorberg: Yes, like the comment says, we explicitly want to verify this.

The problem here was that exceptions that were deserialized would throw NPE when `printStackTrace()` was called on them

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
danielnorberg: Yes.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
honnix: But the lines below uses print writer to verify each line explicitly.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
honnix: Just concerning about unnecessary noise in `System.err`, and it is not clear to me what we are testing here.

danielnorberg: Ok I guess we can remove this loop.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
danielnorberg: Will remove the loop below.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    final Exception deserialized = PersistingContext.deserialize(exceptionFile);
+
+    // Verify that stack trace can be accessed and printed and seems to be correct
+
+    deserialized.printStackTrace();
+
+    final StringWriter sw = new StringWriter();
+    final PrintWriter pw = new PrintWriter(sw);
+    deserialized.printStackTrace(pw);
```
honnix: I mean we do this anyway here so why printing to stdout at line 55?

danielnorberg:
I’m paranoid

honnix:
👍

PR #91, opened 2018-06-14
fix forking for scala 2.11

danielnorberg:
Inspired by https://gist.github.com/retronym/0178c212e4bacffed568#lambdas-in-scala-212-or-211-with-experimental-options

### DIFF
--- a/flo-scala_2.11/pom.xml
+++ b/flo-scala_2.11/pom.xml
@@ -15,10 +15,15 @@
 
   <properties>
     <scala.baseVersion>2.11</scala.baseVersion>
-    <scala.version>2.11.11</scala.version>
+    <scala.version>2.11.12</scala.version>
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.scala-lang.modules</groupId>
+      <artifactId>scala-java8-compat_2.11</artifactId>
+      <version>0.9.0</version>
+    </dependency>
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>flo-workflow</artifactId>
@@ -34,6 +39,11 @@
       <version>${scala.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>flo-runner</artifactId>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>org.scalatest</groupId>
       <artifactId>scalatest_${scala.baseVersion}</artifactId>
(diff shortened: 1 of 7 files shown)


flo-scala_2.12/src/test/scala/com/spotify/flo/DslTest.scala
```diff
@@ -36,5 +39,27 @@ class DslTest extends FlatSpec with Matchers {
     exception should have message "Builder accessor used outside a defTask scope"
   }
 
+  it should "runInProcesses" in {
+    val foo = defTaskNamed("foo")
+      .process(() => {
+        val jvm = ManagementFactory.getRuntimeMXBean.getName
```
danielnorberg: Surprisingly, two processes run but this code seems to be executed in the `bar` process. The `foo` process seems to return a closure, which is then evaluated in the `bar` process.

😕

danielnorberg: Maybe I'm misunderstanding the syntax here...

danielnorberg: nvm my bad, I had obviously declared my process fn to return a closure :shrug:

flo-scala_2.11/pom.xml
```diff
@@ -53,7 +63,11 @@
           <sourceDir>../flo-scala_2.12/src/main/scala</sourceDir>
           <scalaVersion>${scala.version}</scalaVersion>
           <sendJavaToScalac>false</sendJavaToScalac>
-          <addScalacArgs>-Xexperimental</addScalacArgs>
+          <args>
+            <arg>-Xexperimental</arg>
+            <arg>-Ydelambdafy:method</arg>
```
honnix: `delambdafy`, lol

flo-scala_2.12/src/test/scala/com/spotify/flo/DslTest.scala
```diff
@@ -36,5 +39,39 @@ class DslTest extends FlatSpec with Matchers {
...
+      .process({
+        val fooJvm = ManagementFactory.getRuntimeMXBean.getName
+        println(s"foo jvm: $fooJvm")
+        fooJvm
+      })
+    val bar: Task[Array[String]] = defTaskNamed("bar")
+      .input(foo)
+      .context(new TaskContextStrict[String, Array[String]] {
+        override def provide(evalContext: EvalContext): String = {
+          val barContextJvm = ManagementFactory.getRuntimeMXBean.getName
```
honnix: ⭐️

danielnorberg:
I don't entirely understand the root cause of the issue that this resolves though. This just forces a quite different lambda serialization mechanism to be used.

danielnorberg:
We might also need chill-scala if users pass scala things between tasks: https://github.com/twitter/chill#serializers-for-scala-classes

danielnorberg:
https://github.com/spotify/flo/pull/94 might be preferable instead of using experimental compiler flags.

danielnorberg:
Closing in favor of #94

PR #93, opened 2018-06-14
regroup log entry fields

honnix:
@danielnorberg Does this make sense?

I removed those `@Nullable`, not sure how they could be `null`.

### DIFF
--- a/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
+++ b/contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
@@ -22,7 +22,6 @@ package com.spotify.flo.contrib.styx;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.norberg.automatter.AutoMatter;
-import javax.annotation.Nullable;
 
 @AutoMatter
 interface StructuredLogMessage {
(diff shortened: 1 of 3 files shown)


honnix:
@ulzha PTAL as well.

contrib/flo-styx/src/main/java/com/spotify/flo/contrib/styx/StructuredLogMessage.java
```diff
@@ -22,32 +22,38 @@
...
-  @JsonProperty @Nullable String styx_parameter();
-  @JsonProperty @Nullable String styx_execution_id();
-  @JsonProperty @Nullable String styx_trigger_id();
-  @JsonProperty @Nullable String styx_trigger_type();
-  @JsonProperty @Nullable String task_id();
-  @JsonProperty @Nullable String task_name();
-  @JsonProperty @Nullable String task_args();
+  @AutoMatter
+  interface Workflow {
+    @JsonProperty String styx_component_id();
```
danielnorberg: Would it make sense to structure things even more then?

```yaml
{
  "workflow": {
    "styx": {
      "workflow_id":"...",
      "component_id":"...",
      "workflow_id":"...",
      "docker_args":"...",
      "docker_image":"...",
      "commit_sha":"..."
    },
    "framework": "flo|luigi",
    "task": {
      "id": "...",
      "name": "...",
      "args": "..."
    }
  },
  "time": "...",
  "severity": "...",
  "logger": "...",
  "thread": "...",
  "message": "..."
}
```

honnix: Absolutely.

danielnorberg:
@honnix Yeah, those fields can no longer be null (at least not as produced by this code, things might be different if we try to parse some log messages from the future etc).

honnix:
It that ever happens, we can change it safely I think. So far just to get rid of those IntelliJ warnings about potential NPE, which are annoying.

danielnorberg:
yolo!

PR #94, opened 2018-06-14
fix forking for scala 2.11 - redux

danielnorberg:
Alternative to #91 

Instead of using experimental compiler flags, configure kryo to include synthetic fields. 

I think the root cause of our problem was that kryo would drop some synthetic fields in the ScalaApi TaskBuilder & friends. `*waving hands a bit*`

### DIFF
--- a/flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
+++ b/flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
@@ -99,6 +99,7 @@ public class PersistingContext extends ForwardingEvalContext {
     kryo.register(java.lang.invoke.SerializedLambda.class);
     kryo.register(ClosureSerializer.Closure.class, new ClosureSerializer());
     kryo.addDefaultSerializer(java.lang.Throwable.class, new JavaSerializer());
+    kryo.getFieldSerializerConfig().setIgnoreSyntheticFields(false);
 
     if (Files.exists(file)) {
       throw new RuntimeException("File " + file + " already exists");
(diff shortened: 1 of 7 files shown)


flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -99,6 +99,7 @@ public static void serialize(Object object, Path file) throws Exception{
     kryo.register(java.lang.invoke.SerializedLambda.class);
     kryo.register(ClosureSerializer.Closure.class, new ClosureSerializer());
     kryo.addDefaultSerializer(java.lang.Throwable.class, new JavaSerializer());
+    kryo.getFieldSerializerConfig().setIgnoreSyntheticFields(false);
```
danielnorberg: We might not want this to be enabled globally for everything we serialize. We might only want it enabled for 
the TaskBuilder classes and friends in ScalaApi.

danielnorberg: Here's a post by Nate (creator of Kryo) about the [perils of synthetic fields](https://groups.google.com/d/msg/kryo-users/jgAU18bLyN0/h4jEFSGH0WAJ). However, I we are using the exact same JVM when serializing & deserializing, so I think it might be fine to have it enabled globally.

honnix: ok, right. we have full control which jvm to use in this case.

danielnorberg:
About synthetic fields: https://stackoverflow.com/a/5223362

danielnorberg:
I guess this might also handle serialization of closures in user code better. Although I hope users won't be passing closures between tasks...

honnix:
👍

PR #95, opened 2018-06-18
Add ConfigSerializer to PersistingContext

labianchin:
I think I found an issue with flo, kryo and typesafe/lightbend Config serialization:
> "ConfigObject is immutable, you can't call Map.put"
https://github.com/twitter/chill/issues/265

The solution is to create and register a serializer for SimpleConfig class.
This PR adds the serializer on flo-freezer.
Still, I think it might be questionable to add a serializer here upstream. For Config it might generic and worth. But I wonder if later flo should allow users to register their own serializers.

### DIFF
--- a/flo-freezer/pom.xml
+++ b/flo-freezer/pom.xml
@@ -26,5 +26,9 @@
       <groupId>com.twitter</groupId>
       <artifactId>chill-java</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.typesafe</groupId>
+      <artifactId>config</artifactId>
+    </dependency>
   </dependencies>
 </project>
--- a/flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
+++ b/flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
@@ -37,6 +37,8 @@ import com.spotify.flo.TaskId;
 import com.spotify.flo.context.ForwardingEvalContext;
 import com.twitter.chill.java.PackageRegistrar;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.LinkedHashMap;
@@ -127,6 +129,11 @@ public class PersistingContext extends ForwardingEvalContext {
     Kryo kryo = new Kryo();
     PackageRegistrar.all().apply(kryo);
     kryo.register(java.lang.invoke.SerializedLambda.class);
+    try {
+      kryo.register(Class.forName("com.typesafe.config.impl.SimpleConfig"), new ConfigSerializer());
+    } catch (ClassNotFoundException e) {
+      LOG.debug("ConfigSerializer not registered", e);
+    }
     kryo.register(ClosureSerializer.Closure.class, new ClosureSerializer());
     kryo.setInstantiatorStrategy(new Kryo.DefaultInstantiatorStrategy(new StdInstantiatorStrategy()));
     kryo.addDefaultSerializer(Throwable.class, new JavaSerializer());
--- a/flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
+++ b/flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
@@ -95,39 +95,43 @@ public class PersistingContext extends ForwardingEvalContext {
     return promise.value();
   }
 
-  public static void serialize(Object object, Path file) throws Exception{
-    final Kryo kryo = new Kryo();
-    PackageRegistrar.all().apply(kryo);
-    kryo.register(java.lang.invoke.SerializedLambda.class);
-    kryo.register(ClosureSerializer.Closure.class, new ClosureSerializer());
-    kryo.addDefaultSerializer(java.lang.Throwable.class, new JavaSerializer());
-    kryo.getFieldSerializerConfig().setIgnoreSyntheticFields(false);
-
-    if (Files.exists(file)) {
-      throw new RuntimeException("File " + file + " already exists");
-    }
-
+  public static void serialize(Object object, Path file) throws Exception {
     try {
-      try (Output output = new Output(newOutputStream(file, WRITE, CREATE_NEW))) {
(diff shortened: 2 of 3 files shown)


labianchin:
PTAL @danielnorberg @honnix

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -95,39 +97,48 @@ public PersistingContext(Path basePath, EvalContext delegate) {
...
     Kryo kryo = new Kryo();
     PackageRegistrar.all().apply(kryo);
     kryo.register(java.lang.invoke.SerializedLambda.class);
+    try {
+      kryo.register(Class.forName("com.typesafe.config.impl.SimpleConfig"), new ConfigSerializer());
+    } catch (ClassNotFoundException e) {
+      LOG.debug("ConfigSerializer not registered", e);
+    }
     kryo.register(ClosureSerializer.Closure.class, new ClosureSerializer());
     kryo.setInstantiatorStrategy(new Kryo.DefaultInstantiatorStrategy(new StdInstantiatorStrategy()));
```
honnix: It seems we don't set this for serializing. What does it achieve?

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -95,39 +97,48 @@ public PersistingContext(Path basePath, EvalContext delegate) {
...
-      try (Output output = new Output(newOutputStream(file, WRITE, CREATE_NEW))) {
-        kryo.writeClassAndObject(output, object);
-      }
+      serialize(object, Files.newOutputStream(file, WRITE, CREATE_NEW));
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
   }
 
+  public static void serialize(Object object, OutputStream outputStream) {
```
honnix: Any reason we expose another two APIs?

labianchin: Yes. Then we can have unit tests for tasks, trying serialize and deserialize only using Streams and without using the file system.

These unit tests and/or calling `FloRunner.runTask` is important otherwise unserializeable flo tasks would only be noted at runtime in production.

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
labianchin: I am not sure either. I am trying to deduplicate the code a bit. Should we go back having that part duplicated?

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
honnix: 👍

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
honnix: We will take a look later to make sure we don't break anything.

honnix: It doesn't seem to matter because it is only used for deserializing, just being weird. We might do that after https://github.com/spotify/flo/pull/95/files#diff-d7f1d9c331c8fb2c839cdd4ff6cb0a16R121 to make it less confusing.

labianchin: Sure. Shall I do the change?

honnix: Sure please go ahead. You own this PR. =D

labianchin: 👍  Done.

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -95,41 +97,50 @@ public PersistingContext(Path basePath, EvalContext delegate) {
...
-    if (Files.exists(file)) {
-      throw new RuntimeException("File " + file + " already exists");
-    }
-
+  public static void serialize(Object object, Path file) throws Exception {
     try {
-      try (Output output = new Output(newOutputStream(file, WRITE, CREATE_NEW))) {
-        kryo.writeClassAndObject(output, object);
-      }
+      serialize(object, Files.newOutputStream(file, WRITE, CREATE_NEW));
```
honnix: `newOutputStream` has been statically imported, the same for `newInputStream`.

flo-freezer/pom.xml
```diff
@@ -26,5 +26,9 @@
       <groupId>com.twitter</groupId>
       <artifactId>chill-java</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.typesafe</groupId>
```
honnix: This should be `optional` otherwise user code will pull in this dependency anyway and that makes `Class.forName()` meaningless. Is that right? I mean in fact flo-runner depends on typesafe config anyway, right? So maybe no need for `Class.forName()`.

labianchin: The reason for the `Class.forName()` is that `SimpleConfig` is a package private class, so I have not found any other way to reference it. I will add a comment about that.

About having the `config` dependency in `flo-freezer/pom.xml`, that is because the `Config` interface is references in the `ConfigSerializer` class.

honnix: OK now I understood.

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
honnix: @labianchin fix this? i will give +1 then. :)

labianchin: I removed the static imports. Or should I use them?

honnix: Ah, OK, either way. I didn't see that because it didn't close this comment. Sorry.

honnix:
👍

PR #96, opened 2018-06-18
Add travis build

labianchin:
So it runs maven verify on pull requests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+language: java
+
+jdk:
+  - oraclejdk8
+
+cache:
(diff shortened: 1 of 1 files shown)


honnix:
Any particular reason switching to travis from circleci? We do have a plan to fix all circleci issues.

labianchin:
Ahh. Good point. I was looking for PR review builds and had not realized there as a `circle.yaml`.

Travis is being used for DBeam, Featran and Spydra. I understood it might be simpler than circle CI.

Since there is a plan to fix circleci, I will close this.

Issue #103, opened 2018-07-09
Configure forking context with conf file


labianchin:
Currently forking can be disabled with environment variable `FLO_FORKING=false`.

I would be great to have this configuration in typesafe config, like defining something in [`reference.conf`](https://github.com/spotify/flo/blob/master/flo-runner/src/main/resources/reference.conf):

```
flo.forkingEnabled=true
flo.forkingEnalbed=${?FLO_FORKING_ENABLED}
```

rouzwawi:
How about also disabling it by default? Seems to be the cause of a number of bugs lately: #100 #101 #102.

Object serialization and JVM forking is tricky, so we should let `ForkingContext` accumulate some mileage before being enabled by default.

rouzwawi:
Merged #104 now. Seems it missed `0.2.11` though.

danielnorberg:
This has been released. Closing.

PR #104, opened 2018-07-09
Load forking context from config

rouzwawi:
Suggested fix for #103

### DIFF
--- a/flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
+++ b/flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
@@ -26,18 +26,10 @@ import com.spotify.flo.TaskId;
 import com.spotify.flo.Tracing;
 import io.grpc.Context;
 import java.io.IOException;
-import java.lang.management.ManagementFactory;
 import java.util.Collections;
-import java.util.Optional;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * An {@link EvalContext} that runs tasks in sub-processes.
- * <p>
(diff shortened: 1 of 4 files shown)


rouzwawi:
@danielnorberg @honnix Have I understood things correctly here; the way `TracingContext` and `StructuredLoggingEncoder` are set up (with `Context.key("task-id")`) requires the tasks to run in separate JVMs so that other worker-threads do not concurrently also set that same `Context.key`.

danielnorberg:
How about working on fixing the serialization/forking issues instead? We want to fork in order to have per task logging.

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
```diff
@@ -26,55 +26,23 @@
...
-    if (forking.isPresent()) {
-      if (forking.get()) {
-        log.debug("Forking enabled (environment variable FLO_FORKING=true)");
-        return new ForkingContext(baseContext);
-      } else {
-        log.debug("Forking disabled (environment variable FLO_FORKING=true)");
-        return baseContext;
-      }
-    } else if (inDebugger) {
-      log.debug("Debugger detected, forking disabled by default "
```
danielnorberg: I think we'd want to preserve the behavior or disabling forking by default if running in the debugger.

rouzwawi:
@danielnorberg Please have another look. The PR preserves the previous behaviour when in the debugger and has forking enabled by default. But allows it to be configured via a `flo.conf` file in user projects.

danielnorberg:
👍

PR #106, opened 2018-07-10
add NotRetriable (exit code 50) exception

danielnorberg:
Allow a task to signal that it failed and should not be retried.


rouzwawi:
👍

PR #107, opened 2018-07-10
use chill-scala - fix scala serde

danielnorberg:


flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -33,6 +33,8 @@
 import com.spotify.flo.Task;
 import com.spotify.flo.TaskId;
 import com.spotify.flo.context.ForwardingEvalContext;
+import com.twitter.chill.IKryoRegistrar;
+import com.twitter.chill.KryoInstantiator;
```
labianchin: Are we using this? If we are importing `chill` here don't we also have to add `chill` as a dependency on `flo-freezer/pom.xml`?

danielnorberg: We have a direct dependency on chill-java. For chill-scala we rely on flo-scala depending on the appropriate scala version of that.

labianchin: 👍

PR #108, opened 2018-07-17
test mode

danielnorberg:
Flo workflows can be run in a "test mode" without having to modify the existing workflow code:
```java
final Task<Boolean> task = Task.named("task").ofType(Boolean.class)
    .process(() -> {
      // FloTesting.isTest() == true if run in test mode
    });

try (TestScope ts = FloTesting.scope()) {
  // FloTesting.isTest() == true
  FloRunner.runTask(task).future().get();
}
```

`TaskContext` implementations can check whether they are running under test using `FloTesting.isTest()` and change their behavior accordingly. Mock values etc can be stored in the test context using `TestContext.key()` & friends. This is intended for e.g. our internal flo libraries to interact with, not user code directly. Flo libraries should wrap their use of `TestContext` and provide easy to use primitives. E.g.:

```java
try (TestScope ts = FloTesting.scope()) {
  StorageLookup.mock().mockValue("bar", "17");
  
  final Task<String> task = Task.named("task").ofType(String.class)
      .input(StorageLookup.of("bar"))
      .context(new PublishingContext("foo"))
      .process((bar, result) -> {
        // Do some processing and produce a value
        return result.publish(processData(bar));
      });

  assertThat(runTask(task).future().get(30, SECONDS), is("42"));

  assertThat(StorageLookup.mock().lookups("bar"), is(1));
  assertThat(PublishingContext.mock().lookups("foo"), is(1));
  assertThat(PublishingContext.mock().published("foo"), contains("42"));
}
```

In scala:
```scala
flo.test(() => {
  ...
})
```


flo-workflow/src/main/java/com/spotify/flo/TestScope.java
```diff
@@ -0,0 +1,40 @@
...
+
+import io.grpc.Context;
+
+public class TestScope implements AutoCloseable {
+
+  private final Context origContext;
+
+  TestScope() {
+    final TestContext tc = new TestContext();
+    origContext = Context.current().withValue(CONTEXT, tc).attach();
```
rouzwawi: I would prefer to do this work at the call site and pass in the `Context` to this constructor instead. I was a bit confused to what was going on in the methods of `FloTesting` until I read this and noticed the `Context` modifications.

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
```diff
@@ -371,6 +387,137 @@ public void tasksRunInProcesses() throws Exception {
     assertThat(uniqueJvms, not(contains(mainJvm)));
   }
 
+  @Test
+  public void isTestShouldBeTrueInTestScope() throws Exception {
+    try (TestScope ts = FloTesting.scope()) {
+      final Task<Boolean> isTest = Task.named("task").ofType(Boolean.class)
+          .process(FloTesting::isTest);
```
rouzwawi: How does this work?

I might just not understand the grpc Context class enough. It seems like it only has a `ThreadLocalContextStorage` which leads me to think that whatever `TestScope` (with accompanying `TestContext`) would be bound to the test runner thread and not the worker threads.

Does this work because of the propagation in `TracingContext`? If so, maybe there should be an explicit link in the javadoc mentioning the dependency.

danielnorberg: This works because `AsyncContext` [propagates](https://github.com/spotify/flo/blob/master/flo-workflow/src/main/java/com/spotify/flo/context/AsyncContext.java#L47) the grpc `Context` to worker threads.

flo-workflow/src/main/java/com/spotify/flo/TestScope.java
danielnorberg: 👍

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
rouzwawi: Aha, that't pretty neat!

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
```diff
@@ -219,6 +220,12 @@ private EvalContext forkingContext(EvalContext baseContext) {
     final boolean inDebugger = ManagementFactory.getRuntimeMXBean()
         .getInputArguments().stream().anyMatch(s -> s.contains("-agentlib:jdwp"));
 
+    if (FloTesting.isTest()) {
+      // TODO: warn or throw IllegalStateException if config.getBoolean(FLO_FORKING) == true ?
+      LOG.debug("Test run, forking disabled");
```
ulzha: Unsure why it is done...

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
```diff
@@ -371,6 +391,76 @@ public void tasksRunInProcesses() throws Exception {
...
+      assertThat(PublishingContext.mock().published("foo"), contains(jobResult));
+    }
+
+    // Verify that all mocks are cleared when leaving scope
+    try (TestScope ts = FloTesting.scope()) {
+      assertThat(PublishingContext.mock().published("foo"), is(empty()));
+      assertThat(PublishingContext.mock().lookups("foo"), is(0));
+    }
+
+    // Mock a context lookup and verify that the process fn does not run
```
NatashaL: Should this be its own test?

flo-runner/src/test/java/com/spotify/flo/context/Mocks.java
```diff
@@ -0,0 +1,209 @@
...
+import java.net.URI;
+import java.util.ArrayDeque;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@SuppressWarnings("WeakerAccess")
+public class Mocks {
```
NatashaL: It would be nice to expose these mocks to used by the end users as well, aside from the internal tests? Like in a `flo-test` module or as part of the `flo-runner`

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
NatashaL: +1

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
danielnorberg: 👍

flo-runner/src/test/java/com/spotify/flo/context/Mocks.java
danielnorberg: I do not imagine that users will use these mocks of imaginary storage systems. Instead, users will want to mock inputs and outputs for actual production systems. E.g. in our case we want mocks for hades, bigquery, gcs, etc.

flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java
danielnorberg: Commenting

flo-runner/src/test/java/com/spotify/flo/context/Mocks.java
NatashaL: Thanks for clarifying, I had misunderstood the purpose of these mocks.

NatashaL:
👍

PR #111, opened 2018-07-18
Nest exceptions when multiple tasks fail

rouzwawi:
Seemed a bit arbitrary to just silently ignore tasks exceptions in the case multiple tasks fail. This should allow all of them to bubble up as suppressed exceptions.


flo-workflow/src/main/java/com/spotify/flo/Values.java
```diff
@@ -63,6 +64,9 @@ private Values() {
     BiConsumer<T, Throwable> firstComplete = (t, firstThrowable) -> {
       BiConsumer<U, Throwable> secondComplete = (v, secondThrowable) -> {
         if (firstThrowable != null) {
+          if (secondThrowable != null && secondThrowable != firstThrowable) {
+            firstThrowable.addSuppressed(secondThrowable);
```
danielnorberg: if `secondThrowable` is a cause or a suppressed exception of `firstThrowable` we end up with a loop here, but that's ok as `printStackTrace` and loggers etc should handle circular references in exceptions. I'd even remove the `secondThrowable != firstThrowable` check.

flo-workflow/src/test/java/com/spotify/flo/InputSyncCompletionTest.java
```diff
@@ -82,12 +84,61 @@ public void shouldNotCompleteTaskBeforeAllInputsAreDone_AsInputListBefore() thro
...
+        .process((a, b) -> "should not happen");
+
+    AwaitValue<Throwable> eval = new AwaitValue<>();
+    context.evaluate(task).onFail(eval);
+
+    Throwable throwable = eval.awaitAndGet();
+
+    assertThat(throwable, instanceOf(RuntimeException.class));
+    assertThat(throwable.getMessage(), is("failed 2"));
+    assertThat(throwable.getSuppressed().length, is(1));
```
danielnorberg: Check that the expected exception is returned by `getSuppressed`.

flo-workflow/src/test/java/com/spotify/flo/InputSyncCompletionTest.java
```diff
@@ -82,12 +84,61 @@ public void shouldNotCompleteTaskBeforeAllInputsAreDone_AsInputListBefore() thro
...
+        .process((a, b) -> "should not happen");
+
+    AwaitValue<Throwable> eval = new AwaitValue<>();
+    context.evaluate(task).onFail(eval);
+
+    Throwable throwable = eval.awaitAndGet();
+
+    assertThat(throwable, instanceOf(RuntimeException.class));
+    assertThat(throwable.getMessage(), is("failed 3"));
+    assertThat(throwable.getSuppressed().length, is(1));
```
danielnorberg: Same as above.

flo-workflow/src/test/java/com/spotify/flo/InputSyncCompletionTest.java
```diff
@@ -82,12 +84,61 @@ public void shouldNotCompleteTaskBeforeAllInputsAreDone_AsInputListBefore() thro
...
+
+    throwable = throwable.getSuppressed()[0];
+    assertThat(throwable.getMessage(), is("failed 2"));
+    assertThat(throwable.getSuppressed().length, is(1));
```
danielnorberg: Same as above.

danielnorberg:
Looks good apart from some small comments.

flo-workflow/src/main/java/com/spotify/flo/Values.java
rouzwawi: The check is needed since addSuppressed throws if you add an axception to itself. This can happen if two tasks depend on the same task at two different levels.

There was actually a test failure that caught this.

flo-workflow/src/test/java/com/spotify/flo/InputSyncCompletionTest.java
rouzwawi: This is already done in the following lines. Sense for the two other comments above.

flo-workflow/src/main/java/com/spotify/flo/Values.java
rouzwawi: Since the two exceptions should come from different tasks, they should be independent. That’s why I ruled out any causal relationship between them.

The exception could however be the same exception if the same task is depended on in several tasks.

flo-workflow/src/test/java/com/spotify/flo/InputSyncCompletionTest.java
danielnorberg: My bad

PR #113, opened 2018-07-20
forking in test mode

danielnorberg:
\+ Makes it possible to test workflows with forking, catching serialization errors etc at development time instead of in production

\- Makes mocks more complicated
\- Will cause line coverage to not be gathered for tasks ?


NatashaL:
I can't say I understood this PR. @rouzwawi would you mind taking a look? seeing how you've reviewed some other PRs related to forking

PR #114, opened 2018-07-23
scio operator

danielnorberg:


flo-workflow/src/main/java/com/spotify/flo/TaskOperator.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.flo;
+
+public abstract class TaskOperator<T> extends TaskContextGeneric<T> {
```
rouzwawi: Could the job spec type `T` be restricted to some interface that better documents the expected behavior of throwing a `SpecException` on some `success` method?

danielnorberg: Would prefer to come up with a better mechanism instead of abusing exceptions.

rouzwawi: It's a bit ucky, but it's not really exposed to the user so I'm ready to let it slip. Will of course prefer a better solution. Will think a bit about it.

danielnorberg: I think I might retreat a bit and just run the operator directly in the process fn instead.

I realized that executing the operator outside the task process fn does not work well with our current forking strategy as the operator ends up executing on the main process instead of the task process. That might be fixable as well by introducing additional EvalContext methods to allow execution of arbitrary functions to be forked, but feels like the scope of this PR might be becoming too large.

contrib/flo-scio_2.11/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>flo-scala_2.11</artifactId>
+      <version>0.2.14-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>scio-core_2.11</artifactId>
+      <version>0.5.6</version>
```
labianchin: I think this might be hard to keep up to date, as scio/Beam sometimes introduce breaking changes. The risk I see is that whenever there is new beam/scio versions, users of `flo-scio` will have to wait for updates in this lib here.

Have we considered defining `flo-scio` into the internal scio mono repo or maybe as a project here: https://github.com/spotify/scio ?

contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  def pipeline(pipeline: F1[ScioContext, _]): ScioJobSpec[R, S] = {
+    new ScioJobSpec[R, S](taskId, _options, pipeline, _result, _success)
+  }
+
+  def result[RN](result: F2[ScioContext, ScioResult, RN]): ScioJobSpec[RN, S] = {
+    new ScioJobSpec[RN, S](taskId, _options, _pipeline, result, null)
+  }
+
+  def success[SN](success: F1[R, SN]): SN = {
```
NatashaL: The method name here might be a bit confusing. I would call it something like
` def run[SN](onSuccess: F1[R, SN]): SN ` That way it's clear to users that this should be called very last, after setting up the options and pipeline.

This also makes me think that having `run()` and  `success(..)` separate might be even clearer, that would allow adding a `fail(..)` for failure handlers

contrib/flo-scio_2.11/pom.xml
danielnorberg: We're touching a small enough part of the scio/beam api that I'm hopeful breaking changes will be rare.

Moving `flo-scio` into the internal scio mono repo seems like a missed opportunity to have this be open source although the thought has occurred to me as well.

I guess an argument could be made for moving it to the oss `spotify/scio` repo but as flo is still in alpha with much of the api and internals in flux, we'd have the same problem of it being hard to have `spotify/scio` keep up. Also, I fear moving to `spotify/scio` now would slow down the development of `flo-scio`.

contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
danielnorberg: Yeah, the idea was that this job api would declarative and let the operator run the job at a time and place of its choosing. Having a `run()` method makes the api more imperative and implies that the job is immediately executed. Maybe it could be called something else like `done()`. Or maybe it's better to go for the obvious `run()` now and remove it later when/if job execution moves out of the process fn. 🤔

contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+  def options(options: F0[PipelineOptions]): ScioJobSpec[R, S] = {
+    new ScioJobSpec[R, S](taskId, Some(options), _pipeline, _result, _success)
+  }
+
+  def pipeline(pipeline: F1[ScioContext, _]): ScioJobSpec[R, S] = {
+    new ScioJobSpec[R, S](taskId, _options, pipeline, _result, _success)
+  }
+
+  def result[RN](result: F2[ScioContext, ScioResult, RN]): ScioJobSpec[RN, S] = {
+    new ScioJobSpec[RN, S](taskId, _options, _pipeline, result, null)
```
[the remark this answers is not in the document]
danielnorberg: The type signature of `success` is dependent on `result`. But maybe there's some way to work around that with sufficiently smart use of type parameters.

contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
NatashaL: I see, if that was your intention from the beginning, then I understand

contrib/flo-scio_2.11/pom.xml
labianchin: Makes sense. I guess that is something to consider for the future.

contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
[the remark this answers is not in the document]
danielnorberg: Maybe we should just enforce the order of the method calls and set things to null in `options()` and `pipeline()` as well.

NatashaL:
👍

PR #116, opened 2018-07-27
move operator execution of out task process fn

danielnorberg:
* moving blocking execution of out process fn
* prepare for instrumentation to be able to extract the job spec from the task


NatashaL:
nifty 👍

PR #119, opened 2018-07-30
forking: remove redundant task id setup

danielnorberg:
TracingContext is already included in the closure and will set up the task id in the task process.


honnix:
👍

PR #121, opened 2018-07-30
Fix scala type check

NatashaL:
Running mvn scala:doc-jar gave the error:

```
ScioOperator.scala:37: error: overloaded method value key with alternatives:
14:02:22 [INFO]   [T](x$1: String, x$2: com.spotify.flo.TaskBuilder.F0[T])com.spotify.flo.TestContext.Key[T] <and>
14:02:22 [INFO]   [T](x$1: String)com.spotify.flo.TestContext.Key[T]
14:02:22 [INFO]  cannot be applied to (String, () => com.spotify.flo.contrib.scio.ScioOperator.Mocking)
14:02:22 [INFO]   private val MOCK = TestContext.key("mock", () => new Mocking())
```
Explicitly declaring the lambda as a `TaskBuilder.F0` type fixes that

### DIFF
--- a/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala
+++ b/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala
@@ -35,10 +35,9 @@ class ScioOperator extends TaskOperator[ScioJobSpec.Provider] {
 }
 
 object ScioOperator {
-  private val supplier = new F0[Mocking] {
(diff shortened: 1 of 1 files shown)


NatashaL:
@danielnorberg ptal

danielnorberg:
This seems to be due to scaladoc not picking up the `-Xexperimental` argument.

danielnorberg:
I think the real fix is this https://github.com/spotify/flo/pull/122

NatashaL:
Closing in favor of #122

PR #124, opened 2018-07-31
wip: Mock StagingTableId's in flo-bigquery

NatashaL:
A test like 
```
      ScioOperator.mock().jobTest(task.id())(_
        .input(
          BigQueryIO[SpecData.EndContent](
           getQuery(inputTableId)),
          SpecData.fakeInput)
        .output(
          BigQueryIO[SpecData.OutputType](stagingTableId))
        (x =>
          x should containInAnyOrder(SpecData.OutputType))
      )
```

Requires an the `stagingTableId` to be known to the user, cos `.output(..)` seems to get called before `ScioOperator.success` therefore the result data is expected to be in the intermediate staging table.

PR #128, opened 2018-08-02
fix staging table id generation

danielnorberg:
Fix bigquery staging broken in #124 

Reverts the behavior of staging dataset and table generation to be equivalent to before #124.


NatashaL:
👍

Issue #101, opened 2018-07-05
Java properties are not passed to forked task


labianchin:
With the introduction of [ForkingExecutor](flo-runner/src/main/java/com/spotify/flo/context/ForkingExecutor.java#L133), I started noticing some errors on BQ flo tasks.

It seems to be related to the fact that I am running the flo process with `-DGOOGLE_CLOUD_PROJECT=my-gcp-project` and that system property definition is not passed to the the forked task subprocess.

To fix that would require some changes on `ForkingExecutor`, I guess something related to [this](https://stackoverflow.com/a/25966419/1046584).

danielnorberg:
I remember writing the code to propagate properties but maybe it got lost in a rebase. Will fix.

danielnorberg:
Found it: https://github.com/spotify/flo/pull/105

labianchin:
Great!

Still, there can be the case of System properties being configured outside of the command line e.g. with `System.getProperties().setProperty()`. Maybe we don't have to fix for that, but at least acknowledge the limitation in the documentation. WDYT?

danielnorberg:
Maybe system properties and env vars is something that should be configurable per task? Or maybe it’s something that could/should be doable using a task context.

PR #130, opened 2018-08-03
[hold] Propagate user.home to subprocess

NatashaL:
When forking, the `BigQuery` client used by `BigQueryContext` looks for the local google cloud project, which is fetched following this priority of locations: https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L332

The last most alternative is reading from the local `$HOME/.config/gcloud/` and if the preferred home folder location is not propagate properly, the subprocess ends up using the default home folder location https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L370

### DIFF
--- a/flo-runner/src/main/java/com/spotify/flo/context/ForkingExecutor.java
+++ b/flo-runner/src/main/java/com/spotify/flo/context/ForkingExecutor.java
@@ -111,9 +111,10 @@ class ForkingExecutor implements Closeable {
     private final Path resultFile = tempdir.resolve("result");
     private final Path errorFile = tempdir.resolve("error");
 
-    private final String home = System.getProperty("java.home");
+    private final String userHome = System.getProperty("user.home");
+    private final String javaHome = System.getProperty("java.home");
     private final String classPath = System.getProperty("java.class.path");
-    private final Path java = Paths.get(home, "bin", "java").toAbsolutePath().normalize();
+    private final Path java = Paths.get(javaHome, "bin", "java").toAbsolutePath().normalize();
 
     private final Fn<T> f;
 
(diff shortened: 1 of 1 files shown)


NatashaL:
@ulzha @Zatine @labianchin ptal

labianchin:
Makes sense. This seems related to https://github.com/spotify/flo/issues/101 . As a follow up of this PR, I wonder if we should consider passing more/all java properties to the subprocess.

NatashaL:
Ah it seems precisely because checking $HOME is the very last step in trying to get the gcp project, it's still not enough when trying to run on a remote gcp server - it then probably gets the gcp project it's running in after all https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L449

NatashaL:
As @danielnorberg mentions here, maybe this should be configurable per task or per context

danielnorberg:
I don’t understand this. Afaik java defaults the system property`user.home` to be `$HOME` and no manual propagation should be necessary.

If manual propagation is indeed necessary we should have tests verifying the desired effect.

labianchin:
> I don’t understand this. Afaik java defaults the system propertyuser.home to be $HOME and no manual propagation should be necessary.

I think this might be some behavior from distroless java. From experience setting `HOME` env var has no effect on java `user.home` property.

danielnorberg:
But we’re not setting user.home in our code. And afaik distroless doesn’t either. So still doesn’t make sense to me.

NatashaL:
Closing this after verifying flo.forking=false doesn't change the behaviour

PR #132, opened 2018-08-06
more strict check of styx execution id

honnix:
@danielnorberg Do you think this is more hacky by looking into the actual ID?


danielnorberg:
I worry about later when we've forgotten about this and we change the execution id format in styx and structured logging breaks for everyone...

honnix:
It's a valid concern, very unlikely we would change that though. I will give it a second thought.

Issue #118, opened 2018-07-30
flaky test: shouldEvaluate3N_IIL


danielnorberg:
`TaskTest.shouldEvaluate3N_IIL` sometimes fails with the below error:

```
java.lang.AssertionError: timeout while waiting for 0 concurrent tasks to run

	at org.junit.Assert.fail(Assert.java:88)
	at com.spotify.flo.ControlledBlockingContext.waitUntilNumConcurrent(ControlledBlockingContext.java:63)
	at com.spotify.flo.TaskTest.validateEvaluation(TaskTest.java:201)
	at com.spotify.flo.TaskTest.shouldEvaluate3N_IIL(TaskTest.java:150)
```

danielnorberg:
@rouzwawi Do you think this is due to a real issue in how flo executes task graphs or just a bug in the test?

rouzwawi:
@danielnorberg PTAL

PR #134, opened 2018-08-07
Fix race in SyncContext.Value and delegate from ControlledBlockingCon…

rouzwawi:
This fixes the flaky tests reported in #118. I managed to consistently provoke the flakyness by running these tests repeatedly. The fixes below remedied the failures after sufficient runs of the tests (did not run them forever though, so not 100% sure 🤞)

The cause seem to be due to a race in `ControlledBlockingContext.Value` where `consume` and `setValue` could be called from multiple threads and cause the consumer to not be registered/ignored. This issue was also found in `SyncContext` so I fixed it there and refactored `ControlledBlockingContext` to delegate to a `SyncContext`.

This should not be an issue with the `AsyncContext` as it delegates all those methods to an internal `CompletableFuture`.

### DIFF
--- a/flo-workflow/src/main/java/com/spotify/flo/context/SyncContext.java
+++ b/flo-workflow/src/main/java/com/spotify/flo/context/SyncContext.java
@@ -94,12 +94,12 @@ public class SyncContext implements EvalContext {
     }
 
     @Override
-    public void consume(Consumer<T> consumer) {
+    public synchronized void consume(Consumer<T> consumer) {
       valueReceiver.get().accept(consumer);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


rouzwawi:
@danielnorberg merged the `Promise/Value` implementation of `(Sync|Async)Context`. But `SyncContext` uses a same thread executor instead.

rouzwawi:
hmm, unrelated build failure?

```
18:39:49.258 [main] DEBUG com.spotify.flo.freezer.PersistingContext - Could not find com.twitter.chill.ScalaKryoInstantiator: {}
java.lang.ClassNotFoundException: com.twitter.chill.ScalaKryoInstantiator
```

PR #135, opened 2018-08-07
task evaluation logic in EvalContext & let process fn return operation

danielnorberg:
* Remove OperationException
* Remove chaining from task builder
* Allow process fn to return operation spec


flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
```diff
@@ -439,7 +440,7 @@ public void mockingInputsOutputsAndContextShouldBePossibleInTestScope() throws E
             return publisher.publish(result);
           });
 
-      assertThat(runTask(task).future().get(30, SECONDS), is(jobResult));
+      assertThat(runTask(task).future().get(30, DAYS), is(jobResult));
```
rouzwawi: A bit too patient maybe? Would this hang the build if things go wrong?

flo-workflow/src/main/java/com/spotify/flo/TaskOperator.java
```diff
@@ -20,22 +20,15 @@
...
- * {@link OperationException} to be caught and executed by the {@link OperatingContext}.
+ * an operation description. The operation description should be returned from the process fn.
  */
-public abstract class TaskOperator<T> extends TaskContextGeneric<T> {
-
-  public abstract static class OperationException extends ControlException {
-
-    public abstract <T> T run(Listener listener);
-  }
+public abstract class TaskOperator<T, Y, Z> extends TaskContextGeneric<T> {
```
rouzwawi: I noticed a potential pitfall with these type arguments when implementing a `TaskOperator` which lead to user code written using that operator cause a `ClassCastException` even though the code compiled. The changes in `TaskBuilder` do a good job at keeping `T` and `Y` in check since they are used as the input arg type and return type of `process`. However, there is no tie between `Y` and `Z` which means that the type used for `Y` (the jobspec) might be implemented in a way so that it would compile something that return something other than `Z` to `Z perform(Y o, Listener listener)` in `TaskOperator`.

I might not be explaining this very clearly so here's an example using `JobOperator` from the test.

```java
Task<JobResult> task = Task.named("task", today).ofType(JobResult.class)
    .context(JobOperator.create())
    .process(job -> job
        .options(() -> ImmutableMap.of("quux", 17))
        .pipeline(ctx -> ctx.readFrom("foo").map("x + y").writeTo("baz"))
        .validation(result -> {
          if (result.records < 5) {
            throw new AssertionError("Too few records seen!");
          }
        })
        .success(result -> "oh a string!" /* should be JobResult */));
```

This compiles fine, but causes a `ClassCastException`.

I think an interface that would allow `Y` and `Z` to be related could remedy this. But it might restrict how the job spec can be implemented. I'm thinking something like:

```java
public abstract class TaskOperator<T, Y extends OperatorSpec<Z>, Z> extends TaskContextGeneric<T> {

  public interface OperatorSpec<ZZ> {
    ZZ success(/* need freedom here */);
  }

  // ...
}
```

I realise this is mainly a pitfall that operator implementors might run into that then can cause confusion for users. So maybe it can just be documented as something to be aware of in the `TaskOperator` javadoc.

rouzwawi:
👍 💯 on the `TaskBuilder` simplification, it's actually just a builder now 😄 

Had a comment on some type argument interactions in `TaskOperator`. Would be interesting if we could find a way that makes it less fragile.

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
```diff
@@ -67,8 +71,42 @@
...
+      // Run the process fn
+      final Listener listener = listener();
+      final Invokable processFn = task.processFn();
+      final Value<Object> value = context.invokeProcessFn(task.id(), () ->  {
+        final Object result = processFn.invoke(as.toArray());
+
+        // Run operator
+        return operator
+            .map(o -> o.perform(result, listener))
+            .orElse(result);
```
rouzwawi: Does this rely on type checks from `TaskBuilder`? If an operator exists, then `result` will be the same as the `T` type specified by `TaskOperator<T, Y, Z>` and if no operator exists this will be the `T` specified by `Task<T>`? If so, a comment mentioning that would be good here.

flo-api-generator/src/main/resources/TaskBuilderImpl.mustache
```diff
@@ -41,176 +32,127 @@ final class {{implClassName}} {
...
 
     @Override
-    public <A> TaskBuilder1<A, Z> context(TaskContextStrict<A, Z> taskContext) {
+    public <A> TaskBuilder1<A, Z, Z> context(TaskContextStrict<A, Z> taskContext) {
       return context_(taskContext);
     }
 
-    private <A> TaskBuilder1<A, Z> context_(TaskContext<A> taskContext) {
+    @Override
+    public <A, Y> TaskBuilder1<A, Y, Z> context(TaskOperator<A, Y, Z> operator) {
```
rouzwawi: Should this be called `operator`, like in the Scala api?

flo-workflow/src/main/java/com/spotify/flo/TaskOperator.java
rouzwawi: Suggested change: https://github.com/spotify/flo/commit/96dc47042973dba923d871401dd8fdda938a6dbf

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
danielnorberg: Yeah, we rely on type checking happening in the `TaskBuilder` interface.

If an operator exists, result will be `Y` in `TaskOperator<T, Y, Z>`. Will add a comment.

Maybe we should have better names for the type arguments =)

rouzwawi: I was thinking that too now with these types that actually mean something within flo.

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
danielnorberg: woops, debugging leftover =)

flo-workflow/src/main/java/com/spotify/flo/TaskOperator.java
```diff
@@ -28,7 +28,7 @@
  * <p>The concrete operator implementation should {@link #provide(EvalContext)} the task with some means of constructing
  * an operation description. The operation description should be returned from the process fn.
  */
-public abstract class TaskOperator<T, Y extends TaskOperator.OperatorSpec<Z>, Z> extends TaskContextGeneric<T> {
+public abstract class TaskOperator<T, Y extends TaskOperator.OperatorSpec<Z>, Z> implements TaskContext<T> {
```
rouzwawi: Could this be an interface now?

danielnorberg: 👍

danielnorberg:
Is this PR good enough to go?

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
rouzwawi: Missing that comment about the type of result

rouzwawi:
I all looks good, but what would happen if one used two operators? The builder api leads to the last operator type being the expected return type of processFn, but I suspect this would cause problems in `EvalContext` when evaluating the task. @danielnorberg WDYT?

rouzwawi:
I'm fine with fixing the things in the last comment in a separate PR as it probably needs some more thought. 👍 on this PR.

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
danielnorberg: Added a comment

danielnorberg:
@rouzwawi Yes, both `TaskOperator` and `TaskContextStrict` breaks if more than one is used in a task. Let's address that in a later PR.

PR #136, opened 2018-08-08
dry fork when in test mode or debugging

danielnorberg:
* Catch serialization issues when running tests instead of in production


danielnorberg:
@NatashaL ptal

ulzha:
:+1:

PR #139, opened 2018-08-09
read STYX_LOGGING

honnix:
depending on https://github.com/spotify/styx/pull/541


danielnorberg:
LGTM after deploying spotify/styx#541

PR #140, opened 2018-08-10
Scio failure handling

danielnorberg:
Add a method to configure failure handling when using the `ScioOperator`. Example:

```scala
.result((sc, sr) => SomeValidationLibrary.validateCounters(sc, sr)) // throws ValidationError
.success(...)
.failure(t => t match {
  case ValidationError(msg) => throw new NotRetriable(msg) // do not retry on ValidationError
  case _ => throw JobError(t)
})
```

### DIFF
--- a/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
+++ b/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
@@ -28,23 +28,40 @@ class ScioJobSpec[R, S](private[scio] val taskId: TaskId,
                         private[scio] val options: Option[() => PipelineOptions] = None,
                         private[scio] val pipeline: ScioContext => Unit = null,
                         private[scio] val result: (ScioContext, ScioResult) => R = null,
-                        private[scio] val success: R => S = null
+                        private[scio] val success: R => S = null,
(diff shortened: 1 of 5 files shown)


labianchin:
LGTM

But I'd like to have a way to define the same `failure` handling across different pipelines. Maybe we could have a way to specify the failure handling on the ScioOperator constructor/builder/factory?

labianchin:
> Maybe in a separate PR? At least with this PR you can define a shared failure handling method that you can reuse in your tasks.

Merge state


Yes, makes sense.

contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioJobSpec.scala
```diff
@@ -28,23 +28,40 @@ class ScioJobSpec[R, S](private[scio] val taskId: TaskId,
...
+    require(success != null)
+    new ScioJobSpec(taskId, options, pipeline, result, success, failure)
+  }
+
+  def failure(failure: Throwable => S): ScioJobSpec[R, S] = {
+    require(failure != null)
+    new ScioJobSpec(taskId, options, pipeline, result, success, failure)
+  }
+
+  private[scio] def validate(): Unit = {
```
NatashaL: No validation for `options` ?

danielnorberg: 👍

Issue #133, opened 2018-08-07
flo-scio: support for scio 0.5.6 and Beam SDK 2.5.0


labianchin:
https://github.com/spotify/scio/releases/tag/v0.5.6

danielnorberg:
Afaict nothing prevents users from using `v0.5.6`

labianchin:
AFAIK there are incompatible things:

`scio-core:0.5.6` depends on `beam-sdk:2.5.0` which depends on `google-api-client:1.23.0`


`flo-bigquery` depedens on `google-cloud-bigquery:0.21.0-beta` which depends on `google-api-services-bigquery:v2-rev347-1.22.0` which depends on `google-api-client:1.23.0`.

So, time to upgrade `google-cloud-biqquery`?

NatashaL:
> @labianchin:
> this: https://github.com/spotify/flo/blob/master/contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/DefaultBigQueryClient.java#L25 is only compatible with `1.22.0`. `import com.google.cloud.WaitForOption;` does not exist on `1.23.0`

danielnorberg:
The `com.google.cloud.WaitForOption` class is in `google-cloud-core`, not in `google-api-client`.

danielnorberg:
@labianchin I was able to use scio `0.5.6` simply by specifying that version in my `build.sbt`, no flo changes needed. Both the flo and scio BigQuery integrations worked as well.

PR #142, opened 2018-08-13
scio 0.5.6

labianchin:


labianchin:
Not simple as this. Needs upgrade of `google-cloud-bigquery.` See #142 .

danielnorberg:
We still have a users on 0.5.5. I believe users can use 0.5.6 without this change. Am I missing something?

https://github.com/spotify/flo/issues/133#issuecomment-412793087

PR #143, opened 2018-08-13
use circle-ci 2.0

danielnorberg:
circleci 1.0 will be deprecated end of aug.


.circleci/config.yml
```diff
@@ -0,0 +1,52 @@
...
+          paths:
+            - ~/.m2
+      - run:
+          environment:
+            # Configure datastore emulator heap size
+            JAVA: java -Xmx512m -Xms512m
+            # Maven heap size
+            MAVEN_OPTS: -Xmx256m -Xms256m
+          command: mvn -B verify scala:doc-jar -Pcoverage
+      - run: bash <(curl -s https://codecov.io/bash)
```
NatashaL: If I understood this correctly, running this line will simply download the codecov uploader to `bash`, but it's not actually being used for uploading any coverage reports.

.circleci/config.yml
```diff
@@ -0,0 +1,52 @@
...
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - test_jdk8_with_coverage:
+          filters:
+            branches:
+              ignore: gh-pages
+# FIXME
```
NatashaL: What's the issue to be fixed here?

.circleci/config.yml
danielnorberg: It’s calling bash to run the uploader downloaded using curl.

.circleci/config.yml
danielnorberg: Too much for this PR. There is another wip PR for that.

NatashaL:
👍

PR #147, opened 2018-08-15
scio 0.5.6

labianchin:


labianchin:
@danielnorberg @honnix can I have Circle CI on this PR?

honnix:
Strange. It seems @danielnorberg did something to fix circleci already. No idea why there is none for this PR.

labianchin:
> We still have a users on 0.5.5. I believe users can use 0.5.6 without this change.

Right. This PR might not be very important.

But internally on the scio monorepo there isn't support for scio `0.5.5` anymore, only `0.5.6`. I think it is desirable to align all these repos on scio `0.5.6` (desirable, not urgent, not sure if important yet).

PR #149, opened 2018-08-17
scio: report dataflow region, project & url

danielnorberg:


contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala
```diff
@@ -143,10 +147,28 @@ class ScioOperator[T] extends TaskOperator[ScioJobSpec.Provider[T], ScioJobSpec[
...
+    val url = dataflowJobMonitoringPageURL(job)
+    val jobMeta = Map(
+      "job-id" -> job.getJobId,
+      "project-id" -> job.getProjectId,
+      "region" -> job.getRegion,
+      "monitoring-page-url" -> url
+    )
+    log.info("Started scio job (dataflow): {}", jobMeta)
+    val jobMetaJson = new ObjectMapper().writeValueAsString(jobMeta)
+    listener.meta(taskId, "dataflow-job", jobMetaJson)
```
danielnorberg: Went for reporting all of these values in a single atomic piece of metadata as that seems less fragile than sending them as separate metadata values and separate requests to lineage etc.

This could be a little nicer if `listener.meta` would take a `Map<String, String>` instead.

honnix: Let's add that API instead of storing a JSON as string.

honnix: annoying function interface thing... maybe we break the compatibility?

NatashaL:
👍

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
```diff
@@ -144,6 +158,16 @@ public static EvalContext composeWith(EvalContext baseContext, Listener listener
 
   @Override
   public TaskOperator.Listener listener() {
-    return super.listener().composeWith(listener::meta);
+    return super.listener().composeWith(new TaskOperator.Listener() {
```
NatashaL: Difficult to put into words.
This will never be called from `EvalContext`. InstrumentedContext is a `ForwardingEvalContext`, so it forwards the evaluation of a task to another context which is composed within (rather than forwarding to the `super` implementation). According to https://github.com/spotify/flo/blob/master/flo-runner/src/main/java/com/spotify/flo/context/FloRunner.java#L168 that's the base `EvalContext` which is either `SyncContext` or `AsyncContext`. So it forwards the evaluation to another context instance which doesn't contain this listener. Therefore the listener here https://github.com/spotify/flo/blob/master/flo-workflow/src/main/java/com/spotify/flo/EvalContext.java#L92-L103 is simply NOP https://github.com/spotify/flo/blob/master/flo-workflow/src/main/java/com/spotify/flo/EvalContext.java#L180, and we end up calling meta on NOP :/

danielnorberg: Yeah, I guess [this line](https://github.com/spotify/flo/blob/master/flo-workflow/src/main/java/com/spotify/flo/EvalContext.java#L92) should be changed to:

```java
final Listener listener = context.listener();
```

And we should have some tests to cover this. Good catch!

NatashaL: 👍

PR #151, opened 2018-08-20
too many warnings due to unchecked cast

honnix:
PTAL @danielnorberg


flo-api-generator/src/main/resources/TaskBuilderImpl.mustache
```diff
@@ -92,6 +92,7 @@ final class {{implClassName}} {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
```
danielnorberg: I like suppressions to have as small scope as possible. Can we put it just on the offending line?

danielnorberg:
👍

PR #154, opened 2018-08-27
Call correct listener() and add test

NatashaL:
As discussed here https://github.com/spotify/flo/pull/149/files#r212948632


flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
```diff
@@ -89,7 +89,7 @@
           .map(c -> (TaskOperator) c)
           .findAny();
 
-      final Listener listener = listener();
+      final Listener listener = context.listener();
```
danielnorberg: ⭐️

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+import com.spotify.flo.EvalContext;
+import com.spotify.flo.Task;
+import com.spotify.flo.TaskOperator;
+import com.spotify.flo.context.InstrumentedContext.Listener;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+public class EvalContextTest {
+
+  private Listener listener = mock(Listener.class);
```
danielnorberg: Prefer using [`@MockitoJunitRunner`](https://static.javadoc.io/org.mockito/mockito-core/2.6.8/org/mockito/junit/MockitoJUnitRunner.html) and `@Mock` annotations

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+  private TaskOperator operator = mock(TaskOperator.class);
```
danielnorberg: Avoid raw types, use `@Mock` instead.

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+  private Task task = Task.named("test")
```
danielnorberg: Avoid raw types.

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+      .ofType(String.class)
+      .operator(operator)
+      .process((op) -> "result");
+
+  private ArgumentCaptor<TaskOperator.Listener> opListenerCaptor = ArgumentCaptor
```
danielnorberg: Use [`@Captor`](https://static.javadoc.io/org.mockito/mockito-core/2.2.28/org/mockito/Captor.html)

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+      .forClass(TaskOperator.Listener.class);
+
+  @Test
+  public void listenerIsCalled() {
+    final EvalContext syncContext = EvalContext.sync();
+    final EvalContext composedContext = spy(InstrumentedContext.composeWith(syncContext, listener));
```
danielnorberg: This test is about `EvalContext`, no need for `InstrumentedContext` here. Also, [avoid using `spy` if possible](https://static.javadoc.io/org.mockito/mockito-core/1.10.19/org/mockito/Mockito.html#spy).

NatashaL: I tried to mock EvalContext instead of using InstrumentedContext, but it became too painful and unreadable due to so many default methods. So I opted for spying on a `EvalContext.sync()` instance instead.

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,71 @@
...
+    task = Task.named("test")
+        .ofType(String.class)
+        .operator(operator)
+        .process((oprtr) -> oprtr + " result");
+  }
+
+  @Test
+  public void listenerIsCalled() {
+    final EvalContext baseContext = EvalContext.sync();
+    final EvalContext targetContext = spy(EvalContext.sync());
```
honnix: Those default implementation in EvalContext makes mocking frustrating.

honnix: And that is why we spy it instead.

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
```diff
@@ -125,10 +125,27 @@ default void close() throws IOException {}
...
   private final Listener listener;
+  private TaskOperator.Listener taskOperatorListener;
 
   private InstrumentedContext(EvalContext baseContext, Listener listener) {
     super(baseContext);
     this.listener = Objects.requireNonNull(listener);
+    this.taskOperatorListener = super.listener().composeWith(wrapListener(listener));
+  }
+
+  // this is static to avoid capturing "this" in the anonymous class
```
honnix: @NatashaL this is the tricky part. we don't want to pass those contexts to subprocess.

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -0,0 +1,71 @@
...
+    when(targetContext.listener()).thenReturn(listener);
+
+    baseContext.evaluateInternal(task, targetContext).get();
+
+    verify(targetContext, times(1)).listener();
```
honnix: This is equivalent to `verify(targetContext).listener()`.

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
danielnorberg: How about not having this be an anonymous class? Make it a static class instead?

danielnorberg:
Did some simplification, added a test to `FloRunnerTest` that verifies that the listener works at all and added missing propagation in [ChainedListener](https://github.com/spotify/flo/pull/154/commits/dd6e413f9a8f349fcfe75d2db9d369dad2a6c805#diff-6cb662bdd2eef6153d5a62b3c60f249bR64).

flo-workflow/src/main/java/com/spotify/flo/TaskOperator.java
```diff
@@ -39,40 +43,30 @@
...
-          listener.meta(task, key, value);
-        }
-
-        @Override
-        public void meta(TaskId task, Map<String, String> data) {
-          Listener.this.meta(task, data);
-          listener.meta(task, data);
-        }
+      return (task, data) -> {
+        guardedCall(() -> Listener.this.meta(task, data));
```
honnix: If a Listener overrides meta(TaskId, String, String), that would not be captured, right? This is a breaking change anyway, so how about we remove that default method?

danielnorberg: Let's also remove the single-entry for `InstrumentedContext.Listener.meta` method.

honnix: LGTM. There is not something affecting end user.

danielnorberg: Done

flo-runner/src/test/java/com/spotify/flo/context/Jobs.java
```diff
@@ -23,21 +23,25 @@
...
-    private SerializableConsumer<JobResult> resultValidator;
+    private final TaskId taskId;
+
+    private F0<Map<String, ?>> options = Collections::emptyMap;
+    private SerializableConsumer<JobContext> pipelineConfigurator = ctx -> {};
+    private SerializableConsumer<JobResult> resultValidator = result -> {};
     private F1<JobResult, T> successHandler;
 
-    JobSpec() {
+    JobSpec(TaskId taskId) {
```
honnix: Is this change required for this PR? For test purpose maybe?

honnix:
👍

PR #157, opened 2018-08-29
ScioOperator: use spec.options() even on test mode

labianchin:
Why? So that options configuration code paths can be exercised and
validated.


honnix:
Is this a breaking change? @labianchin

honnix:
There should be no case user calling the constructor directly, so should be fine.

PR #161, opened 2018-09-14
remove unused value memoization code

danielnorberg:

## Description
Remove unused memoization code.

## Motivation and Context
This code is not used anywhere.

## Checklist for PR author(s)

- [x] All tests pass

- [x] Risky changes introduced by this PR have been all considered

### DIFF
--- a/flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
+++ b/flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
@@ -559,6 +559,25 @@ public class FloRunnerTest {
     verifyNoMoreInteractions(listener);
   }
 
+  @Test
+  public void evaluatesTaskOnce() throws ExecutionException, InterruptedException {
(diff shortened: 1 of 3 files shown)


honnix:
I think this code was used to dedup tasks and also used by HadesContext internally. Not anymore? Would `ongoing.computeIfAbsent` achieve the same purpose of deduplication?

honnix:
OK I guess not used anymore. I need to refresh my memory.

honnix:
@rouzwawi PTAL I guess it was designed in this way for a purpose (that might not be valid anymore).

rouzwawi:
I believe the initial design was meant to allowing types that were used as return types from a task, to override a task if there was a way to externally determine if the value was already present.

But it turned out it wasn't the best choice to put that on the task type and instead it was moved to `TaskContextStrict` and `OverridingContext`. So even thought this is a breaking change, it should be fine for a new major release.

bergman:
Nice find and cleanup @danielnorberg !

honnix:
👍 to make it official

PR #168, opened 2018-10-02
scio: make scio-test dependency optional

danielnorberg:

## Description
Make the `scio-test` dependency `optional`.

## Motivation and Context
We do not want to include scio-test and its dependencies (scalatest etc) in user production artifacts.

To use `scala-test`, users must now add an explicit dependency. That seems to already be common practice.

Note that the `ScioOperator` code still contains references to classes in `scio-test`, but only in methods that are used as part of testing, like `ScioOperator.mock()`.

Verified that this works both for running a production scio job and local tests using our internal `flo-example`.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/contrib/flo-scio_2.11/pom.xml
+++ b/contrib/flo-scio_2.11/pom.xml
@@ -34,6 +34,7 @@
       <groupId>com.spotify</groupId>
       <artifactId>scio-test_2.11</artifactId>
       <version>0.5.5</version>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>org.apache.beam</groupId>
--- a/contrib/flo-scio_2.12/pom.xml
+++ b/contrib/flo-scio_2.12/pom.xml
@@ -34,6 +34,7 @@
       <groupId>com.spotify</groupId>
       <artifactId>scio-test_2.12</artifactId>
       <version>0.5.5</version>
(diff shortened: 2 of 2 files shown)


contrib/flo-scio_2.11/pom.xml
```diff
@@ -34,6 +34,7 @@
       <groupId>com.spotify</groupId>
       <artifactId>scio-test_2.11</artifactId>
       <version>0.5.5</version>
```
honnix: Not part of this PR, but was it on purpose we don't put `version` in parent pom `dependencyManagement`?

danielnorberg: Probably because that's a scala 2.11 specific dependency.

honnix: For this https://github.com/spotify/flo/pull/168/files#diff-9b2c165d188171b45f6d1e1dcff5b3f5R26, I usually prefer putting it in parent as well, such as https://github.com/spotify/flo/blob/617987f92362d68baaef38b8d0ec831949458b5d/pom.xml#L60

danielnorberg: Compare with https://github.com/spotify/flo/blob/master/flo-scala_2.11/pom.xml#L48

PR #172, opened 2018-10-04
[WIP] hack to get stdout and stderr without subprocess

bergman:

## Description
Weird hack to be able to read stdout and stderr by thread to allow for by-task output.

## Checklist for PR author(s)

- [ ] All tests pass


danielnorberg:
We initially implemented per-task logging using [grpc-context](https://github.com/spotify/flo/pull/82) (which is thread-locals under the hood), but this technique is unable to handle logging and std{err,out} from threads spawned by user code. All user code and libraries using threads would have collaborate in propagation task information to the threads, e.g. by implementing grpc-context propagation, which we [hacked on for scio](https://github.com/spotify/scio/pull/1197). In the end we decided that this approach was impractical and we would fail to capture all log messages and std{err,out} from tasks. In contrast, the now implemented subprocess method was foolproof, easy to reason about and straightforward to implement. 

The forking / process-per-task execution model has the drawback that it imposes serializability constraints on user code, but we argued that this was already the case with dataflow/scio and that users were already to some extent aware and educated about this issue and would be able to deal with it.

Additionally, the process-per-task execution model prepares us for a future managed workflow execution environment where a workflow service orchestrates the flo workflow execution and runs individual tasks in isolated containers with per-task logging, dependency management, metrics, failure handling, retries, etc.

However, this also makes flo itself less suited as a general framework for implementing flow control in an application. I believe this to be a small tradeoff given our focus on flo for orchestration of batch data processing jobs.

bergman:
Thanks for pointing out the pitfalls. Thought you had done something along this path before going for forking. Do you have an idea of how much stdout/stderr logging actually happens compared to slf4j? Because if I understand slf4j correct it should be possible to add metadata to the loggers somehow. It would of course mean the per-task logging only works for frameworks using slf4j.

danielnorberg:
In the case of slf4j it has the same problems wrt threads. Logging calls made from threads that do not have the task information propagated to them (e.g. using grpc-context) would not be possible to attribute to a task.

Issue #170, opened 2018-10-03
Serialization doesn't work with forking


kanterov:
flo version: 0.4.1

Minimal reproducible example:

```java
package com.spotify.data.clickhouse;

import com.spotify.flo.Task;
import com.spotify.flo.context.FloRunner;
import org.apache.beam.sdk.options.PipelineOptions;
import org.apache.beam.sdk.options.PipelineOptionsFactory;

public class App {
  public static void main(String[] args) {
    PipelineOptions opts = PipelineOptionsFactory.fromArgs(args).create();

    FloRunner
        .runTask(task(opts))
        .waitAndExit();
  }

  public static Task<String> task(PipelineOptions opts) {
    return Task.named("task")
        .ofType(String.class)
        .process(opts::toString);
  }
}
```

## Workaround

Create `resources/flo.conf` and put `flo.forking = false` there.

danielnorberg:
Flo uses kryo for serialization and `org.apache.beam.sdk.options.PipelineOptions` is not kryo-serializable.

This can be worked around by instantiating the `PipelineOptions` inside the task, or passing in a closure instead:

```java
import com.spotify.flo.Task;
import com.spotify.flo.TaskBuilder.F0;
import org.apache.beam.sdk.options.PipelineOptions;
import org.apache.beam.sdk.options.PipelineOptionsFactory;

public class App {

  public static void main(String[] args) {
    F0<PipelineOptions> opts = () -> PipelineOptionsFactory.fromArgs(args).create();

    FloRunner
        .runTask(task(opts))
        .waitAndExit();
  }

  public static Task<String> task(F0<PipelineOptions> opts) {
    return Task.named("task")
        .ofType(String.class)
        .process(() -> opts.get().toString());
  }
}
```

danielnorberg:
The difficulty of serializing `PipelineOptions` seems somewhat intentional: 

> {@link PipelineOptions} is intentionally <i>not</i> marked {@link java.io.Serializable}

https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/options/PipelineOptions.java#L182

However, it is json serializable. Would perhaps be good if flo could attempt json serialization for such types.

danielnorberg:
They really don't want `PipelineOptions` to be java serialized: 

> PipelineOptions objects are not serializable

[...]

PR #174, opened 2018-10-05
use jackson for @JsonSerialize annotated types

danielnorberg:

## Description
use jackson serializaion for @JsonSerialize annotated types

## Motivation and Context
Some types (e.g. beam `PipelineOptions`) are not kryo nor java
serializable, but they are jackson serializable.

Added a test for `PipelineOptions`.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/flo-freezer/pom.xml
+++ b/flo-freezer/pom.xml
@@ -30,5 +30,22 @@
       <groupId>com.typesafe</groupId>
       <artifactId>config</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.twitter</groupId>
+      <artifactId>chill_2.11</artifactId>
+      <version>0.9.2</version>
+      <optional>true</optional>
+    </dependency>
+
+    <!--test deps-->
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-direct-java</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/flo-freezer/src/main/java/com/spotify/flo/freezer/JacksonJsonSerializer.java
+++ b/flo-freezer/src/main/java/com/spotify/flo/freezer/JacksonJsonSerializer.java
@@ -76,6 +76,10 @@ class JacksonJsonSerializer extends Serializer {
     if (type.getAnnotation(JsonSerialize.class) != null) {
       return true;
     }
+    final Class superclass = type.getSuperclass();
+    if (superclass != null && hasJsonSerializeAnnotation(superclass)) {
+      return true;
+    }
     for (Class iface : type.getInterfaces()) {
       if (hasJsonSerializeAnnotation(iface)) {
         return true;
--- a/flo-freezer/src/main/java/com/spotify/flo/freezer/JacksonJsonSerializer.java
+++ b/flo-freezer/src/main/java/com/spotify/flo/freezer/JacksonJsonSerializer.java
@@ -1,3 +1,23 @@
+/*-
+ * -\-\-
+ * Flo Freezer
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


danielnorberg:
Addresses #170

danielnorberg:
Not sure why tests fail, works on my laptop! 😄

danielnorberg:
For some reason `scala-maven-plugin` is failing.

ulzha:
Looks cool to me (pun intended... I don't know much about Kryo and what effects `kryo.setRegistrationRequired(false);` will bring, for example).

danielnorberg:
@ulzha kryo registration already defaults to false. I just kept the explicit call because the chill-scala kryo setup was explicitly setting it.

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -127,36 +129,29 @@ public static void serialize(Object object, OutputStream outputStream) {
...
-      }
-      kryo = instantiator.newKryo();
+      kryo = ScalaKryoSetup.getKryo();
     } else {
       // ...for java
       LOG.debug("using chill-java");
-      kryo = new Kryo();
+      kryo = new Kryo() {
+        @Override
+        public Registration getRegistration(Class type) {
```
kanterov: It isn't completely valid to override `getRegistration`, because it looks more like default serializer. There is `getDefaultSerializerForAnnotatedType` to override.

kanterov: For instance, this will make not possible to register custom serializers for anything that `JacksonJsonSerializer` can comprehend.

danielnorberg: Unfortunately Kryo [special handles](https://github.com/EsotericSoftware/kryo/blob/master/src/com/esotericsoftware/kryo/Kryo.java#L486) `Proxy` instances in a way which means that `getDefaultSerializerForAnnotatedType` can't be used to handle `PipelineOptions`.

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -180,4 +176,32 @@ public static String cleanForFilename(TaskId taskId) {
...
+    try {
+      ScalaKryoInstantiator.class.getName();
+      return true;
+    } catch (NoClassDefFoundError e) {
+      LOG.debug("Could not find com.twitter.chill.ScalaKryoInstantiator: {}", e.toString());
+      return false;
+    }
+  }
+
+  private static class ScalaKryoSetup {
```
honnix: What's the benefit having a static class here instead of just a static method? And why is the branch `using chill-java` inlined instead of being factored out?

flo-freezer/pom.xml
```diff
@@ -30,5 +30,22 @@
...
       <artifactId>config</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.twitter</groupId>
+      <artifactId>chill_2.11</artifactId>
+      <version>0.9.2</version>
```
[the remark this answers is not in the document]
danielnorberg: It's an awkward dependency to have in the parent pom because then we should move the chill dependencies from the scala submodules as well into the parent pom and then we end up enumerating the scala versions in the parent pom... do we want that?

flo-freezer/src/main/java/com/spotify/flo/freezer/JacksonJsonSerializer.java
```diff
@@ -0,0 +1,97 @@
...
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  static Optional<Registration> getRegistration(Kryo kryo, Class type) {
+    return hasJsonSerializeAnnotation(type)
+        ? Optional.of(kryo.getRegistration(RegistrationMarker.class))
+        : Optional.empty();
+  }
+
+  static void register(Kryo kryo) {
```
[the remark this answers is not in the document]
danielnorberg: You have nominated yourself for a nitpick award ;P

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
danielnorberg: It needs to be a static class because we don't want the method to be class-loaded before we know that we have chill-scala available.

danielnorberg: https://github.com/spotify/flo/pull/174/files#diff-d7f1d9c331c8fb2c839cdd4ff6cb0a16R135

flo-freezer/pom.xml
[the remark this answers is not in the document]
honnix: Ok then.

honnix:
Let's close this in favour of #178 ?

PR #178, opened 2018-10-10
use java serialization

danielnorberg:

## Description
Use Java serialization instead of Kryo.

## Motivation and Context
* Many classes and libraries used by users are java-[`Serializable`](https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html) but not kryo-serializable (i.e. needs custom serializer)
* The java `Serializable` mechanism offers a straightforward way for users to customize how their objects are serialized, which is something we miss today with the way we use kryo.
* Google cloud dataflow uses java serialization (for DoFn, closures etc), which is similar to our use case and given our overlap in users we might be able to expect that users are educated on how to work with it.

Tests pass.

## Checklist for PR author(s)

- [x] All tests pass

- [x] This PR has been incorporated in release note for the coming version
- [x] Risky changes introduced by this PR have been all considered


flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
```diff
@@ -129,17 +129,11 @@ public static EvalContext composeWith(EvalContext baseContext, Listener listener
 
   @Override
   public <T> Value<T> invokeProcessFn(TaskId taskId, Fn<T> processFn) {
-    return delegate.invokeProcessFn(taskId, () -> {
-      listener.status(taskId, Listener.Phase.START);
-      try {
-        final T value = processFn.get();
-        listener.status(taskId, Listener.Phase.SUCCESS);
-        return value;
-      } catch (Throwable t) {
```
kanterov: Why there is no try-catch anymore? Is it safe to call `invokeProcessFn`?

flo-runner/src/main/java/com/spotify/flo/context/LoggingContext.java
```diff
@@ -51,17 +51,11 @@ static EvalContext composeWith(EvalContext baseContext, Logging logging) {
...
   @Override
   public <T> Value<T> invokeProcessFn(TaskId taskId, Fn<T> processFn) {
-    return super.invokeProcessFn(taskId, () -> {
-      logging.startEval(taskId);
-      final long t0 = System.nanoTime();
-      try {
-        final T tValue = processFn.get();
-        logging.completedValue(taskId, tValue, Duration.ofNanos(System.nanoTime() - t0));
-        return tValue;
-      } catch (Throwable e) {
```
kanterov: the same comment regarding try-catch

flo-runner/src/main/java/com/spotify/flo/context/Logging.java
```diff
@@ -28,17 +28,21 @@
...
 import com.spotify.flo.status.TaskStatusException;
+import java.io.Serializable;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import org.slf4j.Logger;
 
-class Logging {
+class Logging implements Serializable {
```
kanterov: In beam, they solve this problem by using memoizable suppliers (from guava) that are serializable. I would rather refactor the code to use them, then making `Logging` serializable.

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
honnix: The try-catch was around `processFn` and with the change we don't call `get` on it directly and let `value.consume/onFail` handle it.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
```diff
@@ -71,6 +80,29 @@ public void exceptionSerialization() throws Exception {
     assertThat(deserialized.getStackTrace()[0].getMethodName(), is("exceptionSerialization"));
   }
 
+  @Test
+  public void serializeShouldPropagateSerializationExceptions() {
+    final Object o = new Object();
```
honnix: Let's inline this. `o` is not used anywhere else.

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
```diff
@@ -71,6 +80,29 @@ public void exceptionSerialization() throws Exception {
...
+    exception.expect(RuntimeException.class);
+    exception.expectCause(instanceOf(NotSerializableException.class));
+    PersistingContext.serialize(o, new ByteArrayOutputStream());
+  }
+
+  @Test
+  public void serializeShouldPropagateIOException() throws Exception {
+    final Object o = new Object();
```
honnix: Same here.

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
```diff
@@ -165,21 +164,21 @@ public void nonBlockingRunnerDoesNotBlock() throws Exception {
...
+    assertThat(happenedFile.exists(), is(false));
 
-    Files.write(latchFile, new byte[0]);
+    latchFile.createNewFile();
   }
 
   @Test
   public void blockingRunnerBlocks() throws IOException {
-    final Path file = temporaryFolder.newFile().toPath();
+    final File file = temporaryFolder.newFile();
```
honnix: This change doesn't seem to be necessary. Maybe just to be consistent with other file usages?

flo-scala_2.11/pom.xml
```diff
@@ -33,11 +33,6 @@
       <artifactId>scala-library</artifactId>
       <version>${scala.version}</version>
     </dependency>
-    <dependency>
-      <groupId>com.twitter</groupId>
-      <artifactId>chill_2.11</artifactId>
```
honnix: ⭐️

flo-workflow/src/main/java/com/spotify/flo/BaseRefs.java
```diff
@@ -20,21 +20,35 @@
...
+  private static final long serialVersionUID = 1L;
 
   final Fn<List<Task<?>>> inputs;
   final List<TaskContext<?, ? super Z>> taskContexts;
   final List<ProcessFnArg> args;
   final TaskId taskId;
   protected final Class<Z> type;
 
+  /**
+   * No-arg constructor for serialization
```
honnix: This is for deserialization of child class to be precise, right?

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
```diff
@@ -84,10 +84,10 @@
       // Call preRun on task contexts after all args are evaluated
       task.contexts().forEach(tc -> tc.preRun(task));
 
-      final Optional<TaskOperator> operator = task.contexts().stream()
+      final TaskOperator operator = task.contexts().stream()
```
honnix: Why this change?

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
```diff
@@ -84,10 +84,10 @@
...
           .filter(c -> c instanceof TaskOperator)
           .map(c -> (TaskOperator) c)
-          .findAny();
+          .findAny().orElse(null);
```
honnix: If the change is needed, let's put `orElse` on a new line.

kanterov:
I tried this PR with our project, and it worked. I had to make several things serializable that come from our internal libraries, going to submit PRs there. One potential improvement would be to check that closures are serializable in builders, so it gives a proper stack trace that can be easily debugged. Nevertheless, each exception included what class isn't serializable, that is a huge improvement.

kanterov:
One clarification, it worked without closure cleaner.

flo-workflow/src/main/java/com/spotify/flo/context/InstrumentedContext.java
honnix: Gets propagated all the way up and that fails evaluation of the task, I think. Can check it later.

danielnorberg: Calling `invokeProcessFn` is "safe" as in if it throws an unchecked exception that is either a jvm failure or a flo internal failure and we let that failure propagate all the way to the program entry point, i.e. "panic" and crash. The call to `invokeProcessFn` will not fail due to exceptions thrown by user code.

flo-runner/src/main/java/com/spotify/flo/context/LoggingContext.java
danielnorberg: Same as https://github.com/spotify/flo/pull/178/files/46b674bec5553ff3e1ebc3fdfbb6ef526bbbf201#r224669023

flo-runner/src/test/java/com/spotify/flo/context/FloRunnerTest.java
danielnorberg: `Path` is not serializable.

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
danielnorberg: `j.u.Optional` is not serializable

danielnorberg:
@kanterov checking for serializability in the builders is a good idea 👍

flo-runner/src/main/java/com/spotify/flo/context/Logging.java
danielnorberg: It was a cheap hack to put stateful task status tracking into the `Logging` class to begin with. Refactoring.

flo-workflow/src/main/java/com/spotify/flo/EvalContext.java
danielnorberg: Some interesting discussion about this: https://stackoverflow.com/a/24564612

Gist of it is that `Optional<>` was never meant to be stored as a field. I sympathize with that but seems they lost that war because _everyone_ is doing that. Surprisingly unpragmatic decision.

danielnorberg:
@kanterov I'm thinking that if Beam can get away without a closure cleaner in their library, then we can as well. But perhaps this is naive and we really need one because we offer scala bindings and scalac is more prone to synthesize unnecessary scope references in closures? 

If it currently works for you without a closure cleaner I think we can wait and add that in a later PR if necessary.

honnix:
Not incorporating a closure cleaner in this PR sounds like a good idea to me.

honnix:
This PR LGTM. It should be 0.5.0 then.

kanterov:
I tried different closures with Scala 2.12, and it works well without ClosureCleaner in most of the cases that can happen. Didn't try 2.11 though.

kanterov:
It should work much better in Scala in general because Spark community has been pushing different libraries to make things serializable for a long time, as well as case classes are serializable by default.

kanterov:
👍

rouzwawi:
👍

honnix:
@danielnorberg Please remove the blocking labels and merge it.

PR #179, opened 2018-10-12
task builder: require java serializability

danielnorberg:

## Description
Require java serializability at task build time.

Continuation of https://github.com/spotify/flo/pull/178

## Motivation and Context
Fail early to ease users serialization debugging.

Added tests.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/flo-api-generator/src/main/resources/TaskBuilderImpl.mustache
+++ b/flo-api-generator/src/main/resources/TaskBuilderImpl.mustache
@@ -13,6 +13,7 @@ import static com.spotify.flo.BuilderUtils.inputArg;
 import static com.spotify.flo.BuilderUtils.inputsArg;
 import static com.spotify.flo.BuilderUtils.lazyFlatten;
 import static com.spotify.flo.BuilderUtils.lazyList;
+import static com.spotify.flo.BuilderUtils.requireSerializable;
 import static com.spotify.flo.Values.toValueList;
 
 /**
@@ -37,22 +38,26 @@ final class {{implClassName}} {
 
     @Override
     public Task<Z> process(F0<Z> f) {
+      requireSerializable(f, "process fn");
       final Invokable code = a -> f.get();
       return Task.create(inputs, taskContexts, type, taskId, code, args);
     }
 
     @Override
     public <A> TaskBuilder1<A, Z, Z> context(TaskContextGeneric<A> context) {
+      requireSerializable(context, "context");
       return context_(context);
     }
 
     @Override
     public <A> TaskBuilder1<A, Z, Z> output(TaskOutput<A, Z> output) {
+      requireSerializable(output, "output");
       return context_(output);
     }
 
     @Override
     public <A, Y> TaskBuilder1<A, Y, Z> operator(TaskOperator<A, Y, Z> operator) {
+      requireSerializable(operator, "operator");
       return context_(operator);
     }
 
@@ -64,6 +69,7 @@ final class {{implClassName}} {
 
     @Override
     public <A> {{interfaceName}}1<A, Z, Z> input(Fn<Task<A>> aTask) {
+      requireSerializable(aTask, "input");
       final Fn<Task<A>> aTaskSingleton = Singleton.create(aTask);
       return new Builder1<>(
           lazyFlatten(inputs, lazyList(aTaskSingleton)),
@@ -72,6 +78,7 @@ final class {{implClassName}} {
 
     @Override
     public <A> {{interfaceName}}1<List<A>, Z, Z> inputs(Fn<List<Task<A>>> aTasks) {
+      requireSerializable(aTasks, "inputs");
       Fn<List<Task<A>>> aTasksSingleton = Singleton.create(aTasks);
       return new Builder1<>(
           lazyFlatten(inputs, lazyFlatten(aTasksSingleton)),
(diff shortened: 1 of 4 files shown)


kanterov:
There is a bit of more happening in Beam implementation that checks for serialization:
- https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/util/SerializableUtils.java#L78

kanterov:
What about including taskId in exception message?

flo-freezer/src/test/java/com/spotify/flo/freezer/PersistingContextTest.java
```diff
@@ -101,6 +105,46 @@ public void deserializeShouldPropagateSerializationExceptions() {
     PersistingContext.deserialize(new ByteArrayInputStream("foobar".getBytes()));
   }
 
+  @Test
+  public void shouldProvideDetailedDebugInformationOnSerializationExceptions() throws ReflectiveOperationException {
```
danielnorberg: TODO: Move serialization code and tests from `flo-freezer` module to `flo-workflow` to avoid duplication of code. Separate PR.

danielnorberg: https://github.com/spotify/flo/pull/182

flo-freezer/src/main/java/com/spotify/flo/freezer/PersistingContext.java
```diff
@@ -52,6 +52,11 @@
  */
 public class PersistingContext extends ForwardingEvalContext {
 
+  static {
+    // Best effort. Hope that ObjectOutputStream has not been loaded yet :pray:
+    System.setProperty("sun.io.serialization.extendedDebugInfo", "true");
```
kanterov: Can we add this option when we do forking?

danielnorberg: Done.

flo-runner/src/test/java/com/spotify/flo/context/OverridingContextTest.java
```diff
@@ -60,11 +60,13 @@ public String provide(EvalContext evalContext) {
     }
   };
 
-  int countUpstreamRuns = 0;
-  int countRootRuns = 0;
+  private static int countUpstreamRuns = 0;
```
honnix: It's kind of subtle to understand why they have to be `static` unfortunately. Might be worth commenting a bit.

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
```diff
@@ -39,16 +41,16 @@
   private static final String SPEC = "spec";
   private static final String RESULT = "result";
 
-  @Mock private TaskOperator.Listener listener;
-  @Mock private TaskOperator<String, String, String> operator;
+  @Mock private static TaskOperator.Listener listener;
```
honnix: This doesn't need to be `static`.

flo-workflow/src/test/java/com/spotify/flo/context/InstrumentedContextTest.java
```diff
@@ -86,19 +86,19 @@ public void callsStatusForFailingTask() throws Exception {
     ));
   }
 
-  Task<String> upstream(int i) {
+  private static Task<String> upstream(int i) {
```
honnix: Did these test cases just succeed accidentally? If I understood correctly, the primary change in this PR is to enforce a roundtrip of serialisation.

flo-workflow/src/test/java/com/spotify/flo/ForwardingTaskContextGeneric.java
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.flo;
+
+public class ForwardingTaskContextGeneric<T> extends TaskContextGeneric<T> {
```
honnix: Do we remember why `TaskContextGeneric` was not an interface to begin with?

danielnorberg: I do not remember. I believe we should try to make it an interface, but that's not in scope for this PR.

flo-workflow/src/test/java/com/spotify/flo/context/EvalContextTest.java
danielnorberg: 👍

danielnorberg:
Wrt the complexity of serialization, apart from allowing per-task logging I also saw it as something that would enable managed execution of Flo workflows by a workflow service that orchestrates execution at the task level and spawns a container per task, persists intermediate results, runs dataflow jobs asynchronously etc.

And yes, there is clearly some friction between the (now incorrectly) intuited shared memory _programming model_ where tasks can have tight out-of-band coupling and pass/hold references to shared mutable state in the same process etc, and the actual _execution model_ we're implementing where tasks are isolated from each other and only interact via explicitly declared inputs and outputs.

flo-workflow/src/test/java/com/spotify/flo/context/InstrumentedContextTest.java
honnix: Yeah that is what I meant as "accidentally".  When designing the test tasks we didn't always have serializable in mind. Enforcement is good!

PR #182, opened 2018-10-19
move serialization code to `flo-workflow` module

danielnorberg:

## Description
move serialization code to `flo-workflow` module

## Motivation and Context
Avoid code duplication.

## Checklist for PR author(s)

- [ ] All tests pass


flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
```diff
@@ -80,19 +82,25 @@ static EvalContext dryComposeWith(EvalContext baseContext) {
     LOG.debug("Test run, forking disabled - testing serialization");
```
honnix: Let's remove this debug log because is function will be called from different paths: dry or test, where we already log.

flo-workflow/src/main/java/com/spotify/flo/BuilderUtils.java
```diff
@@ -41,11 +41,6 @@
  */
 class BuilderUtils {
```
honnix: We can add more test cases for this class to get better coverage.

flo-runner/src/main/java/com/spotify/flo/context/ForkingContext.java
danielnorberg: 👍

flo-workflow/src/main/java/com/spotify/flo/BuilderUtils.java
danielnorberg: Separate PR?

flo-runner/src/main/java/com/spotify/flo/context/ForkingExecutor.java
```diff
@@ -129,7 +130,7 @@ void start() {
       }
       log.debug("serializing closure");
       try {
-        PersistingContext.serialize(f, closureFile);
+        Serialization.serialize(f, closureFile);
       } catch (Exception e) {
```
honnix: Is it intended to catch all kinds of exceptions here? And similar code below.

danielnorberg: Oops

honnix:
👍

PR #192, opened 2018-11-20
Add lookupLatestDaily method to BigQueryTasks

benkogan:

## Description

This PR adds a lookupLatestDaily method to BigQueryTasks for looking up a date-partitioned (by table name) BigQuery table starting from a given date with a look-back day limit.

## Motivation and Context

Motivated by actual need. Use case is a pipeline that depends on a date-partitioned table but has a certain bound of late data tolerance and so should proceed with an older table rather than fail on the absence of a table for a single static date.

I have included unit tests. I also ran a simple task using lookupLatestDaily via FloRunner and tested correct behavior for table presence/absence.

## Checklist for PR author(s)

- [x] All tests pass
- [ ] Relevant documentation updated (there were no existing docs around this so left as-is)
- [x] This PR has NO breaking change to public API

### DIFF
--- a/contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/BigQueryLookup.java
+++ b/contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/BigQueryLookup.java
@@ -21,7 +21,9 @@
 package com.spotify.flo.contrib.bigquery;
 
 import com.google.cloud.bigquery.TableId;
+import com.spotify.flo.util.Date;
 import java.io.Serializable;
+import java.util.Objects;
 import java.util.Optional;
 
 class BigQueryLookup implements Serializable {
@@ -33,6 +35,10 @@ class BigQueryLookup implements Serializable {
 
   Spec lookup(TableId tableId) {
     return new LookupSpec(tableId);
+  }
+
+  Spec lookupLatestDaily(String project, String dataset, String tableName, Date start, int lookBackDays) {
+    return new LookupLatestDailySpec(project, dataset, tableName, start, lookBackDays);
   }
 
   interface Spec extends Serializable {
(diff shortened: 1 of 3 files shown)


contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/BigQueryLookup.java
```diff
@@ -55,4 +60,40 @@ Spec lookup(TableId tableId) {
...
+    private static final long serialVersionUID = 1L;
+
+    final String project;
+    final String dataset;
+    final String tableName;
+    final Date start;
+    final int lookBackDays;
+
+    LookupLatestDailySpec(String project, String dataset, String tableName, Date start, final int lookBackDays) {
+      this.project = project;
```
danielnorberg: null checks

contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/BigQueryTasks.java
```diff
@@ -50,6 +49,17 @@ private BigQueryTasks() {
     return lookup(TableId.of(project, dataset, table));
   }
 
+  public static Task<TableId> lookupLatestDaily(F0<FloBigQueryClient> bigQuerySupplier, String project, String dataset, String tableName, Date start, final int lookBackDays) {
```
danielnorberg: package-private

contrib/flo-bigquery/src/main/java/com/spotify/flo/contrib/bigquery/BigQueryLookup.java
benkogan: should they be added to only the public-facing API in BigQueryTasks or to both that and the constructor here as well?

benkogan: went ahead with the former but lmk if that's the wrong approach and will swap!

danielnorberg:
Nice!

All tests pass, merging.

PR #198, opened 2018-12-10
Manage flo-tests version

NatashaL:

## Motivation and Context

Fix patching version for `flo-tests-shading` and `flo-tests-shading-user`. Seems you can't use properties for setting the artifact version dynamically.

Work around https://issues.apache.org/jira/browse/SUREFIRE-1588 by using a slight older base image.

## Checklist for PR author(s)

- [ ] All tests pass

### DIFF
--- a/flo-tests/shading-user/pom.xml
+++ b/flo-tests/shading-user/pom.xml
@@ -10,7 +10,7 @@
 
   <name>Flo Tests - Shading User</name>
   <artifactId>flo-tests-shading-user</artifactId>
-  <version>${flo.version}</version>
+  <version>placeholder-version</version>
   <description>
     Tests using shaded flo
   </description>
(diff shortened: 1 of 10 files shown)


flo-tests/shading-user/pom.xml
```diff
@@ -10,7 +10,7 @@
 
   <name>Flo Tests - Shading User</name>
   <artifactId>flo-tests-shading-user</artifactId>
-  <version>${flo.version}</version>
+  <version>placeholder-version</version>
```
danielnorberg: We don't even need this to be the "real" version. It can be any version we want. E.g. `0.0.1-SNAPSHOT` or whatever.

danielnorberg: Not sure this is causing the build to fail though.

PR #212, opened 2020-02-03
Update to scio 0.8 and other dependencies that required version update

aleksandr-spotify:

## Description

Update to scio 0.8 + other dependencies update, when needed.
## Checklist for PR author(s)

- [x] All tests pass
- [x] Relevant documentation updated
- [x] This PR has breaking change to public API and it is documented

### DIFF
--- a/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala
+++ b/contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala
@@ -67,15 +67,15 @@ class ScioOperator[T] extends TaskOperator[ScioJobSpec.Provider[T], ScioJobSpec[
       spec.pipeline(sc)
 
       // Start job
-      val scioResult = Try(sc.run())
-      scioResult match {
+      val sec = Try(sc.run())
+      sec match {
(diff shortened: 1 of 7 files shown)


contrib/flo-scio_2.12/src/main/scala/com/spotify/flo/contrib/scio/ScioOperator.scala
```diff
@@ -67,7 +67,7 @@ class ScioOperator[T] extends TaskOperator[ScioJobSpec.Provider[T], ScioJobSpec[
       spec.pipeline(sc)
 
       // Start job
-      val scioResult = Try(sc.close())
+      val scioResult = Try(sc.run())
```
honnix: Does this mean we will stop supporting old versions of Scio?

honnix: So `.close()` is not yet removed from Scio https://github.com/spotify/scio/blob/v0.8.1/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L678

honnix: It is not compatible at byte code level due to method signature change.

